PR #45, opened 2017-07-25
added --rand-nonce option

Fusl:
The code by default tries nonces in a sequential order starting from a fixed/calculated value.

This option allows to override this behavior by trying random instead of increasing nonces to help mining with poorly developed mining pools that hand out the exact same job to all users.


J-kit:
Thanks m8, i'd love to see this in the official repo :-)

xmrig:
Can you provide link to that pools?
Also I don't sure it correct method to create random nonces? For example https://github.com/JayDDee/cpuminer-opt/blob/master/cpu-miner.c#L1513
It looks better, randomize only start nonce, cares about thread count, and sequential after that.

Fusl:
I'm sorry, but I can't provide a URL to the pool I am using as it is a private pool.

Using my code doesn't have much overhead in general and makes sure that all hashes are generated pseudo-randomly while the [cpuminer-opt](https://github.com/JayDDee/cpuminer-opt/blob/master/cpu-miner.c#L1513) code generates this hash when starting a job and then starts incrementing from there on, which means, if two or more miners in the network create the same hash, the share the same nonce and therefore get rejected. A Pseudo-RNG for every hash is on the efficiency level still very high and generates perfectly fine random numbers to be used for hashing and it's impossible that two or more clients generate the same nonce and therefore the same hash.

xmrig:
I need think about it, I don't like use rand after each hash and I think cpuminer-opt method enough in this case too, but maybe I'm wrong. Your private pool is come kind of proxy? otherwise it bad pool. If yes I think nicehash method for proxy much better and predictable. Thank you.

Fusl:
How about using rand to shift the nonce every 256 hashes? This should almost completely eliminate the inefficiency factor of rand and still allow the miner to produce good, unique nonces andhashes.

xmrig:
Sounds better. And my question again that pool is a proxy?

Fusl:
Yes, as far as I am aware, the pool is a proxy to http://xmrpool.eu/

Fusl:
@xmrig Please check my last commit

J-kit:
xmrig --print-time=15 --max-cpu-usage=100 -t 7 -a cryptonight -o stratum+tcp://j-kit.de:3334 -u WeLoveBurgers -p x

xmrig:
Okay guys I will check it in next 24 hours, now is little bit busy.

In addition, I will release new proxy after some time, still no ETA, written on c++/libuv, but I use nicehash nonce method (up to 256 miners to one pool connection). Extremely high performance, can easy handle over 100k connection on 1024 MB virtual machine, I used benchmark for check it.

J-kit:
Any news? When will this be merged into the repo?

PR #87, opened 2017-09-04
Add support to GCC Version less than 7.1

draskolnikova:
Added compile support, gcc less than 7.1 didn't support `-std=c++14`.


xmrig:
Please check this commit https://github.com/xmrig/xmrig/commit/80b0ac2492ef7c26304526fa08b3e063de0e28ef in dev branch, should be better solution.

gcc 5.4 supports `-std=c++14` anyway now features from this standard not used (was used before).

draskolnikova:
@xmrig yes it should be more optimized than checking `c++` version using if. 
Many thanks.

PR #131, opened 2017-09-29
Add stdio header for linux builds

forvelin:
Currently cmake builds on linux fail without those headers. This
commit fixes that.


xmrig:
I recently merge dev branch into master, these headers already added.
Any way what exactly Linux distributive/version did you use? On Ubuntu 16 no issues with these headers
Thank you.

PR #155, opened 2017-10-14
Add Dockerfile

tobwiens:
Add Dockerfile


xmrig:
Looks good for me, but personally I don't use docker at all, so can't verify it.
@calvintam236 can you please verify it?

calvintam236:
Looks like it will work. However, docker image created by this Dockerfile will be huge - there isn't any image size optimization at all.

@tobwiens maybe you should take a look what I did to optimize the size in my Dockerfile for `xmrig`. https://hub.docker.com/r/calvintam236/xmrig/

rtau:
Just for your information, in addition to the contribution of @calvintam236 and @tobwiens, there are several other attempts to run xmrig on docker <https://hub.docker.com/search/?isAutomated=0&isOfficial=0&page=2&pullCount=0&q=xmrig&starCount=0>.

Among them, some of the mostly used docker image are https://hub.docker.com/r/bitnn/alpine-xmrig/ and https://hub.docker.com/r/cryptohash/xmrig-docker/, which are both around 5MB large only by using Alpine instead of Debian/Ubuntu, which seems to be quite common for Docker images.

Please note that cryptohash/xmrig-docker is patching the code before building the Docker image for some gain......

tobwiens:
@calvintam236 Your Dockerfile looks more appropriate in formatting and size. Mine was a Saturday night hack to get started ASAP. 

@rtau I agree those images are pretty small in total.  Usually I use Ubuntu based images so it's in total smaller for me. 

Yes cryptohash/xmrig-docker changes the code to compile with a donate-level to 0 (constexpr const int kDonateLevel = 0) And the command has hard-coded wallet address it seems. So not a good fit.

When you are looking for optimization then I recommend using @calvintam236 file ( because you cleanup after the build ``` && rm -r xmrig \
    && apt-get -y remove ca-certificates git build-essential cmake \
    && apt-get -y autoremove \
    && apt-get clean autoclean \
    && rm -rf /var/lib/{apt,dpkg,cache,log}``` ) 

With an combination of an Alpine fixed v3.6 version (So that future builds don't break the container). 

I update my PR accordingly.

tobwiens:
@xmrig @calvintam236  @rtau  On a second thought that is exactly what @mEcHaPiLoT has written in the Dockerfile (even with cleanup). 

You can remove the sed command which sets the donate-level to 0 and you can change the ENTRYPOINT. 

Maybe @mEcHaPiLoT can create an PR, or give us permission to use the Dockerfile.

andypost:
I used this to compile latest (2.4.3) with Alpine 3.7, final image size *10.9MB* 
Needs to move

``` Dockerfile
FROM  alpine:3.7
RUN set -ex && \
  addgroup -S miner && \
  adduser -S -D -h /xmrig -G miner miner && \
  apk --no-cache upgrade && \
  apk --no-cache add \
    libuv \
    libmicrohttpd && \
  apk --no-cache add --virtual .build-deps \
    git \
    cmake \
    libuv-dev \
    libmicrohttpd-dev \
    build-base && \
  cd xmrig && \
  git clone https://github.com/xmrig/xmrig build && \
  cd build && \
  cmake -DCMAKE_BUILD_TYPE=Release . && \
  make && \
  cd .. && \
  cp build/xmrig /usr/bin && \
  rm -rf build && \
  apk del .build-deps || return 0

USER miner
WORKDIR    /xmrig
ENTRYPOINT  ["/usr/bin/xmrig"]
```

tobwiens:
@andypost Thank you, that's much more optimised. Can you can open a new PR and link it to this one.

andypost:
@tobwiens it makes sense to use docker to build new releases this way & package them to automated builds to docker hub

z3dm4n:
Instead of hard coding the version of Alpine, you should consider using `FROM alpine:latest`. 
You should also consider to delete the line `apk --no-cache add libuv libmicrohttpd`, because whether libuv nor libmicrohttpd packages are needed to run or build xmrig. 

I like this dockerfile [here](https://hub.docker.com/r/bitnn/alpine-xmrig/~/dockerfile/). This file only lacks the libmicrohttpd-dev package or the flag that disables the build support for libmicrohttpd (-DWITH_HTTPD=OFF).

andypost:
If you need to remove this libs then cmake needs to compile with this libs statically linked right now it throws If I remove `libuv`
```
/xmrig # xmrig 
Error loading shared library libuv.so.1: No such file or directory (needed by /usr/bin/xmrig)
```

z3dm4n:
You're right. If you don't want a statically linked xmrig binary, sure you need libuv (and optionally libmicrohttpd). I got confused on this extra step in your dockerfile. But I like the way you solved the removal of the unneeded build dependencies after the build via the `--virtual .build-deps` flag. I didn't know that apk has such a nice feature like this. 

In my dockerfile I disabled libmicrohttpd at build time to further improve security, but this is of course up to you.

sandys:
hi guys,
are the docker builds not being updated ? I'm wondering where to pull from after the hard fork

rpherrera:
Anyone interested in another version (capable of disabling httpd) might want to take a look at:
https://github.com/rpherrera/xmrig-docker-build/blob/master/xmrig-cpu/Dockerfile

Issue #64, opened 2017-08-18
error compiling FileLog.cpp


prismspecs:
```
1>------ Build started: Project: xmrig, Configuration: Debug x64 ------
1>FileLog.cpp
1>C:\Users\dagobah\Downloads\xmrig\src\log\FileLog.cpp(37): error C2065: 'O_CREAT': undeclared identifier
1>C:\Users\dagobah\Downloads\xmrig\src\log\FileLog.cpp(37): error C2065: 'O_APPEND': undeclared identifier
1>C:\Users\dagobah\Downloads\xmrig\src\log\FileLog.cpp(37): error C2065: 'O_WRONLY': undeclared identifier
1>C:\Users\dagobah\Downloads\xmrig\src\log\FileLog.cpp(37): error C2660: 'uv_fs_open': function does not take 5 arguments
1>C:\Users\dagobah\Downloads\xmrig\src\log\FileLog.cpp(91): warning C4267: 'argument': conversion from 'size_t' to 'unsigned int', possible loss of data
1>Done building project "xmrig.vcxproj" -- FAILED.
2>------ Skipped Build: Project: ALL_BUILD, Configuration: Debug x64 ------
2>Project not selected to build for this solution configuration 
========== Build: 0 succeeded, 1 failed, 3 up-to-date, 1 skipped ==========

```

NmxMilk:
Hello,
You should provide context information (paltform, compiler version, etc.) and steps to reproduce your problem !
At first glance, this seems to be due to missing include files (man open under linux) :
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>

The error concerning uv_fs_open may indicate you're using the wrong version of libuv.
Hope this helps
Cheers

xmrig:
According #65 issue you try use mingw library, you should use **libuv.lib** not **libuv.a**.
Now you can use prebuilt libuv https://github.com/xmrig/xmrig-deps/releases I receive a lot of questions about where find libuv.
Also you need change Debug build to Release.
Thank you.

prismspecs:
Thank you. So with the pre built uvlib, I should re-create the build folder and then execute the cmake line in command line, referencing these files rather than program files/uvlib/ etc?

xmrig:
Right.

prismspecs:
Alright, getting closer! Now I've got:

```
[...]

Issue #35, opened 2017-07-11
MSYS64 build


Nivoyash:
Searched everything, but can not find the libuv.a file to build. I found the old version 1.8 but on it I get the error:
`hustle@home MINGW64 /c/Users/hustle/Downloads/xmrig-master/xmrig-master
$ mkdir build3

hustle@home MINGW64 /c/Users/hustle/Downloads/xmrig-master/xmrig-master
$ cd build3

hustle@home MINGW64 /c/Users/hustle/Downloads/xmrig-master/xmrig-master/build3
$ cmake .. -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release -DUV_INCLUDE_DIR="C:\libuv-master\include" -DUV_LIBRARY="C:\libuv-master\lib\x64\libuv.a"
-- The C compiler identification is GNU 7.1.0
-- The CXX compiler identification is GNU 7.1.0
-- Check for working C compiler: C:/msys64/mingw64/bin/cc.exe
-- Check for working C compiler: C:/msys64/mingw64/bin/cc.exe -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: C:/msys64/mingw64/bin/c++.exe
-- Check for working CXX compiler: C:/msys64/mingw64/bin/c++.exe -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found UV: C:/libuv-master/lib/x64/libuv.a
-- Looking for syslog.h
-- Looking for syslog.h - not found
-- Configuring done
-- Generating done
-- Build files have been written to: C:/Users/hustle/Downloads/xmrig-master/xmrig-master/build3

hustle@home MINGW64 /c/Users/hustle/Downloads/xmrig-master/xmrig-master/build3
$ make
Scanning dependencies of target jansson
[  1%] Building C object src/3rdparty/jansson/CMakeFiles/jansson.dir/dump.c.obj
[  3%] Building C object src/3rdparty/jansson/CMakeFiles/jansson.dir/error.c.obj
[  5%] Building C object src/3rdparty/jansson/CMakeFiles/jansson.dir/hashtable.c.obj
[  7%] Building C object src/3rdparty/jansson/CMakeFiles/jansson.dir/hashtable_seed.c.obj
[...]

PR #205, opened 2017-11-18
Fixing failed build on MSVC 2017 Win64

JLospinoso:
* Fixing cmake MHD_INCLUDE_DIR to correspond with README
* Adding missing headers to Options/FileLog to compile on MSVC 2017 Win64


xmrig:
I get 2 issues with FileLog #35 #64 before, both unclear, possible wrong libuv version used.
Thank you.

PR #324, opened 2018-01-06
can build without microhttpd when WITH_HTTPD=OFF

stanz2g:
remove microhttpd from target_link_libraries when WITH_HTTPD=OFF


xmrig:
Thank you.

PR #385, opened 2018-02-04
improved performance with non-AES CPU and Intel L2 cache size fix

Foudge:


Foudge:
@2010phenix You confirm that xmrig **still** detect correct L2/L3 cache ? If yes, it's a good news.
I don't understand why you use av=3, your CPU has AES instructions. It's just for test?
Original code use essentially SSE2 instruction : maybe SandyBridge (codename of Core iX 2xxx family) have better affinity with them, contrary to Core 2 Duo/Quad. But no importance because they prefer av=1.
You seems to have other CPU : have you some results (hashrate and/or L2 size)?

xmrig:
Thank you I merged your changes.

2010phenix:
@Foudge it's for test that nothing break down )

sergneo:
@Foudge is working on the exp version or is it stopped? what progress? at least tell me what's the news

Foudge:
Hi, I am still working on it ;)
But it's getting longer and the gains are getting smaller. The "reward" is inversely proportional to the time spent :/
There is actually no significal performance difference compared to "exp" version compiled with GCC.

sergneo:
@Foudge 
Well, maybe this is the limit for this processor, finish the work with the source for xmrig 👍  :)

Foudge:
The limit I'm going to reach, it's mine, not the CPU ;)

jovan2009:
@Foudge 
The modifications that you made are applicable also for arm processors?
On my [Lenovo a2010](https://www.gsmarena.com/lenovo_a2010-7507.php) I get about 7.5-8.1 H/s with current  xmrig from github. I managed to compile xmrig in termux with clang. With gcc-7 I get an error: `-/data/data/com.termux/files/usr/bin/ld: cannot find -lc++_shared`.
The highest H/s my phone can do so far is 9.2-9.5 H/s with [NeoNeonMiner](https://play.google.com/store/apps/details?id=com.kangaderoo.neoneonminer) but it isn't compatible with xmrig-proxy, no --nicehash option. I wonder if there is room for improvement with xmrig.

Foudge:
@jovan2009 No, only for x86 processors, and only those who do not have AES-NI instructions.
I have no knowledge about ARM processors.
You really do mining on your phone or it's just test or curiosity?

jovan2009:
@Foudge I am LOL. Started as a curiosity but I have 2 identical phones laying around and they give combined almost as much as my core2 duo laptop (well, before you optimized the x86 code :-) ). I know I gain almost nothing as financial value but I learned a lot in the process, like for example how to compile from source in termux, so it's a win.

cryptonius:
Thanks for fixing this! :)

I'd love it if you took a look at the Radeon 5770 issue...

lisergey:
I've just tested new release, supposedly with your improvements, @Foudge.

XMRig/2.4.5 libuv/1.19.1 clang/3.4.1
CPU:          Intel(R) Core(TM) i3 CPU         530  @ 2.93GHz (1) x64 -AES-NI
CPU L2/L3:    0.5 MB/4.0 MB

+4-5 h/s, 6%.

Great job! Thank you!!

auto-joe:
@jovan2009 see my project here which compiles xmrig 2.4.5 natively on Raspbian (rpi3b) with gcc-7.3.0: https://github.com/auto-joe/rPi-xmrig-gcc7.3.0
You might be able to follow what I did on my pi for the ARM OS/Arch you're interested in.

PR #409, opened 2018-02-21
New wiki page: Arch Build

f1she3:
Hello, i think it would be pretty easy to include build instructions for Arch Linux to the Wiki :
# Arch Linux
```
sudo pacman -S libmicrohttpd
git clone https://github.com/xmrig/xmrig.git
cd xmrig
mkdir build
cd build
cmake ..
make
```
# Additional CMake options

- `-DWITH_LIBCPUID=OFF` Disable libcpuid. Auto configuration of CPU after this will be very limited.

- `-DWITH_AEON=OFF` Disable CryptoNight-Lite support.

- `-DWITH_HTTPD=OFF` Build without built in http server and API.

- `-DUV_LIBRARY=/usr/lib/x86_64-linux-gnu/libuv.a` Use static libuv version.


k0ste:
FYI: [aur/xmrig](https://aur.archlinux.org/packages/xmrig)

f1she3:
Ok, right :)

PR #426, opened 2018-03-03
Changes for the Monero v1 PoW

vtnerd:
Changes for the monero version pow changes. Some implementation notes:

- The arm version needs to be tested.
- All versions access the scratchpad directly after the first tweak, instead of unpacking the simd register. The 11 byte needs to be manipulated, as you can see.
- All versions modify the 64-bit stack value instead of the scratchpad.
- A boolean template argument was added to remove branching from the aeon version.
- The version parameter isn't technically needed (the callee can get this from the input), but I decided to do this because all the other implementations were doing this - in case the miner is used for monero-like variants it can be configured at runtime. The penalty is setting up a register for the call.

### DIFF
--- a/src/crypto/CryptoNight.cpp
+++ b/src/crypto/CryptoNight.cpp
@@ -24,6 +24,34 @@
 
 #include "crypto/CryptoNight.h"
 
+// VARIANT ALTERATIONS
+#define VARIANT1_CHECK() \
+    if (MONERO && version > 6 && size < 43) \
+    { \
+        return false; \
+    }
+
+#define VARIANT1_INIT(part) \
+    const uint64_t tweak1_2_##part =\
+        (MONERO && version > 6) ? \
+            (*reinterpret_cast<const uint64_t*>(reinterpret_cast<const uint8_t*>(input) + 35 + part * size) ^ \
+            *(reinterpret_cast<const uint64_t*>(ctx->state##part) + 24)) : 0;
+
+#define VARIANT1_1(p) \
+    do if(version > 6) \
+    { \
+        const uint8_t tmp = reinterpret_cast<const uint8_t*>(p)[11]; \
+        static const uint32_t table = 0x75310; \
+        const uint8_t index = (((tmp >> 3) & 6) | (tmp & 1)) << 1; \
+        ((uint8_t*)(p))[11] = tmp ^ ((table >> index) & 0x30); \
+  } while(0)
+
+#define VARIANT1_2(p, part) \
+    do if(MONERO && version > 6) { \
+        (p) ^= tweak1_2_##part ; \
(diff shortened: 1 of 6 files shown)


xmrig:
Thank you for pull request. I will review changes and merge it by hand. ARM (at least ARMv8) work fine.

src/crypto/CryptoNight.cpp
```diff
@@ -24,6 +24,34 @@
...
+    }
+
+#define VARIANT1_INIT(part) \
+    const uint64_t tweak1_2_##part =\
+        (MONERO && version > 6) ? \
+            (*reinterpret_cast<const uint64_t*>(reinterpret_cast<const uint8_t*>(input) + 35 + part * size) ^ \
+            *(reinterpret_cast<const uint64_t*>(ctx->state##part) + 24)) : 0;
+
+#define VARIANT1_1(p) \
+    do if(version > 6) \
```
stoffu: Shouldn't this be `MONERO && version > 6` like other places?

stoffu:
@vtnerd 
Aeon is also going to need to adopt the same POW tweak from version 2 (https://www.reddit.com/r/Aeon/comments/83rcdh/psa_pow_change_and_testnet_v2_fork_on_35000_march/). As such, could you please modify the logic from:

```c++
if (MONERO && version > 6) { ... }
```

to

```c++
if ((MONERO && version > 6) || (!MONERO && version > 1) { ... }
```
?
Thank you.

xmrig:
@stoffu This pull request currently merged in dev branch, but looks like need some rewrite, remove MONERO template argument at least. I will do it soon.
Thank you.

src/crypto/CryptoNight.cpp
xmrig: Current code there https://github.com/xmrig/xmrig/blob/dev/src/crypto/CryptoNight_monero.h

stoffu:
@xmrig 
Thank you!

Aeon has not yet decided on the fork height, but it will likely be (quite a bit) later than Monero's fork on 28th of March. So I think there's no need for hurry.

stoffu:
@vtnerd @xmrig 

To reduce the chance of possible mistakes in writing branches for the tweak, Aeon plans to bump its block version from v1 to v7 at once which is the same as Monero's upcoming block version. The desired change for the branch thus would be from `if (MONERO && version > 6) { ... }` to simply `if (version > 6) { ... }`.

xmrig:
I already implemented Aeon support in dev branch, to avoid version issues I use version independent `variant` (0 original algo, 1 new). Version is not big issue for Aeon, because only one coin use -lite algo, but good for simplify code. Other coins must use v7 to avoid compatibility issues.
Thank you.

PR #484, opened 2018-03-29
Bit shifting

vodkar:
This code causes undefined behavior if the N value is higher than 32. You cannot use this code to write a value higher than 0x80000000 into the 'X' variable.


Rezacb0413 (approved):

vodkar:
Hello! How about to accept this PR? 😅
I can create a new one if you need it.

SChernykh:
I'm sorry, but this PR makes no sense. The bit shift there is only 3, not 32. Also, this is a 3rd-party code which was taken almost unchanged from another library.

vodkar:
Okey. I'll close it then

PR #514, opened 2018-04-07
Add default file function

lzc-China:
Miners are involved in leasing, so request this feature.
For example: I lease 10 miners for 24 hours and need to switch to the mine pool and address provided by the user, and after the lease time (24 hours) arrives, load the default profile and restart the miners.
We hope to add this function on the WEB interface, and display the status and countdown of the rental, as well as user information.
And change the screening mode set by multiple miners to directly enter all configuration commands, using the filter mode is prone to errors
If completed, a certain amount of XMR donations will be provided
Thanks!


xmrig:
Wrong pull request.

PR #550, opened 2018-04-13
add scheduler in XmRigCCServer

ivoryblade:
Could you please add "scheduler" in XmRigCCServer so that one can set time when each miner should start and stop mining.


xmrig:
Wrong pull request.

PR #570, opened 2018-04-20
Valgrind showed conditonal jumps when printing the hashrate

AngeGallego:
Not much to be said, just had to initialize these strings


2010phenix:
Thanks for sharing.
maybe with this mess over monero v7 and other initialize lost?

xmrig:
Actually it's safe to leave these buffers not initialized, `snprintf` will add terminating `\0` character. So after changes valgrind is ok?
Thank you.

AngeGallego:
Oh indeed it's `snprintf`. Well yes Valgrind is not aware of this and is throwing the warning, which this fixed.

PR #629, opened 2018-05-14
Fix FileLog (--log-file) fails on non-seekable files

Low-power:
Using option `--log-file` on non-seekable files such as pipe or tty, will fail; due to using offset `0` for `uv_fs_write(3)`. This patch fixed the bug by setting offset to `-1`, which is means don't change offset.

### DIFF
--- a/src/common/log/FileLog.cpp
+++ b/src/common/log/FileLog.cpp
@@ -93,5 +93,5 @@ void FileLog::write(char *data, size_t size)
     uv_fs_t *req = new uv_fs_t;
     req->data = buf.base;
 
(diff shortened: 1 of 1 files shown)


xmrig:
Can you show example, how reproduce the bug.
Thank you.

Low-power:
To test, create a pipe `mkfifo /tmp/test-pipe`, then `cat /tmp/test-pipe` from a terminal; running `xmrig --log-file /tmp/test-pipe ...` from another terminal; `cat(1)` will got EOF and exit.
The expecting behaver is the log messages will be transfered via the pipe, and printed out from `cat(1)`

xmrig:
Merged. Thank you.

PR #636, opened 2018-05-17
Changing Nonce Changing

mowoe:
If you are mining on a pool and mining on multiple devices, the Hash for one nonce comes multiple times as the miners are increasing their nonce by one and the nonce is equal on every miner so the hash from the slower miners get rejected. If every nonce is randomly changed, the hashes wont get rejected as the hashes are different from each miner.


mowoe:
i was a bit unclear, i mean if you are mining with multiple miners on one proxy.

PR #657, opened 2018-05-29
reupdate workers.cpp

snipeTR:
add header and line 278>>>288


snipeTR:
XMrig is pushing my computer too hard.
Taskmgr.exe does not even run fast.
I made a change for it.

PR #659, opened 2018-05-29
Create ExeTrigger.cpp

snipeTR:
this func need.
XMrig is pushing my computer too hard.
Taskmgr.exe does not even run fast.
I made a change for it.


tarris034:
This is retarded. Please don't add it.

You can adjust the intensity in config.

Issue #618, opened 2018-05-08
Algo balancing


MoneroOcean:
To be able to do algo load balancing from pool side I suggest the following changes to xmrig* miners and xmrig-proxy:

1. Miner sends list of expected algorithm hashrates in `algo-perf` parameter ("cn", "cn-lite", "cn-heavy", "cn-fast" algo names can be used):

```
{
  "id": 1, "jsonrpc": "2.0", "method": "login",
  "params": {
    "login": "...", "pass": "...", "agent": "...",
    "algo": ["cn", "cn/0", "cn/1", "cn/xtl", "cn-heavy", "cn-light/1" ],
    "algo-perf": {"cn": 600.0, "cn-heavy": 400.0 }
  }
}
```

2. Miner config can include extended `threads` section that describes thread setup for "cn", "cn-lite", "cn-heavy" algo names:

```
"threads": {
	"cn": [
		{"low_power_mode": true,  "affine_to_cpu": 0 },
	        {"low_power_mode": false, "affine_to_cpu": 1 },
	        {"low_power_mode": 1,     "affine_to_cpu": 2 },
	        {"low_power_mode": 3,     "affine_to_cpu": false }
	],
	"cn-lite": [
		{"low_power_mode": true,  "affine_to_cpu": 0 },
	        {"low_power_mode": false, "affine_to_cpu": 1 },
	        {"low_power_mode": 1,     "affine_to_cpu": 2 },
	        {"low_power_mode": 3,     "affine_to_cpu": false }
	],
	"cn-heavy": [
		{"low_power_mode": true,  "affine_to_cpu": 0 },
	        {"low_power_mode": false, "affine_to_cpu": 1 },
	        {"low_power_mode": 1,     "affine_to_cpu": 2 },
	        {"low_power_mode": 3,     "affine_to_cpu": false }
	]
}
```

3. Miner config includes optional extra `algo-perf` section that describes expected hashrates for each algorithm ("cn", "cn-lite", "cn-heavy", "cn-fast" algo names can be used). By default, this section is included in config with estimated generic values (absolute values do not matter, only their relation to each other):

```
"algo-perf": {
	"cn": 1000.0,
	"cn-fast": 2000.0,
	"cn-lite": 2000.0,
	"cn-heavy": 700.0
}
```

[...]

PR #719, opened 2018-07-18
Algo balancing

MoneroOcean:
This is suggested implementation of https://github.com/xmrig/xmrig/issues/618 (only in CPU xmrig for now, not in xmrig-proxy)

In short it do the following:

1) Adds --calibrate-algo (functionality that measured relevant algo hashrates) and --save-config (that can save algo hashrates into algo-perf section of config.json file) switches.
2) Adds extended "threads" option that allows to specify separate thread profiles for different algorithms.
3) All possible algorithms are now reported by miner to the pool (not only variants of the the same base Algo).
4) Allows miner to change algorithm to any supported one specified by "algo" pool job switch.

P.S. The most serious workaround was done to pass config object (with algo hashrates) to Client object via global config object pointer (pconfig).

P.P.S. I have temporary test pool that supports algo/algo-perf miner login if required.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,7 @@ set(HEADERS
     src/net/strategies/DonateStrategy.h
     src/Summary.h
     src/version.h
+    src/workers/Benchmark.h
     src/workers/CpuThread.h
     src/workers/Handle.h
     src/workers/Hashrate.h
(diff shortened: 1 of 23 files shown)


bobbieltd:
Very cool ! +1 ! Vote for MO !

PR #721, opened 2018-07-22
Algo balancing

MoneroOcean:
Replaces PR #719


MoneroOcean:
I added public fr.moneroocean.stream (of [https://moneroocean.stream](https://moneroocean.stream) pool) that can be used for testing of this PR (currently RYO is supported for cn-heavy algo and Masari is supported for cn-fast algo). You should specify XMR wallet address in any case since pool trades all coins to XMR to do payments.

PR #745, opened 2018-08-29
Cryptonight variant 2 support

SChernykh:
Reference code is here: https://github.com/monero-project/monero/pull/4218
I've tested this on x86-64 and ARM for all possible combinations of "algo" and "av" parameters in config.json and it passed self-tests. Performance for variants 0&1 didn't change, compiler even generates the same assembly code (except it uses different registers). Performance for variant 2 is the same as in my repository for tests and benchmarks.


xmrig:
Thank you for pull request. Is any test pool available right now?

SChernykh:
There is no test pool now because Monero's PR is not merged yet. It should be merged this week (hopefully) and then Monero's testnet will switch to variant 2.

xmrig:
I merge your commits into [feature-cn-variant2](https://github.com/xmrig/xmrig/commits/feature-cn-variant2) branch and will create same branches for other projects.

I added 32 bit support for x86 also I got invalid hashes on MSYS2 gcc 7.1 (lazy to update it on my dev machine) https://github.com/xmrig/xmrig/commit/075565e6fa4f346c90ef6cb02028d52ab6363a64 but gcc 7.2+ works fine, it might be my local issue, need additional checks.
Thank you.

SChernykh:
Yes, this subborrow/addcarry bug was probably only fixed in gcc 7.2

2010phenix:
@xmrig Hi.
if monero move to variant 2 maybe need and for classic miner some love? maybe remove variant 0..
I can test when be test pool

xmrig:
@2010phenix sure.

2010phenix:
Thx @xmrig.
I think we have some data for v8 already: https://github.com/monero-project/monero/pull/4358/commits/7f590e72619e81b943af9ad48858472ea78c0918

xmrig:
Merged, all new information/discussion will be placed there #753

PR #761, opened 2018-09-22
Cryptonight variant 2 - final version

SChernykh:
Reference code: https://github.com/monero-project/monero/pull/4404

I tested it on x86 with av=1-10 and on ARM with av=1-4, self test passed.


xmrig:
Merged. Thank you.

PR #764, opened 2018-09-24
Fix for MSYS/MINGW

SChernykh:
Always use Windows calling convention for asm code when running on Windows.


xmrig:
I fix it by different way, I think separated `cnv2_main_loop_win.S` file for Windows is more clear way.
Thank you.

PR #766, opened 2018-09-28
Update Asm.cpp

Roboto12:
parse(value.IsBool()) should be changed to parse(value.GetBool()), otherwise if the 'asm' parameter in config has a boolean value, it will be parse as 'true' even if it's actually 'false'.


xmrig:
Thank you, somehow missed this PR.

PR #834, opened 2018-10-21
Added asm optimized code for AMD Bulldozer

SChernykh:
Tested on on dual Opteron 6276 server (32 cores, 32 MB cache, 24 threads with auto config): one minute average hashrate went up from 701.2 H/s to 717.0 H/s (+2.25%).


xmrig:
Looks good, will merge tomorrow.
Thank you.

PR #848, opened 2018-10-25
Update ConfigLoader_platform.h

snipeTR:
Update help -threads


xmrig:
`"all"` value was created by a friend request for some weak i3 CPUs, in general, use all available threads is bad idea, I don't like special attention for this value, it can cause a lot of question and myths, like `max-cpu-usage` was.
Thank you.

PR #887, opened 2018-12-02
Add UPX Support

QuantumLeaper:
Pull request to add uPlexa (UPX) support. Iteration sequence of (2^18) / 2
https://github.com/uPlexa/uplexa


xmrig:
* Algorithm name should follow this rule `base algorithm/variant`, so it should named `cryptonight-lite/upx`, if follow this rule it also reduce number of changed lines.
* Also you should add test values to this file https://github.com/xmrig/xmrig/blob/master/src/crypto/CryptoNight_test.h

I close this PR, please create new PR with required changes against dev branch.
Thank you.

PR #934, opened 2019-02-17
Add support for Intel C Compiler

minhng99:
I have a minor hashrate increase with ICC, my core i3-7100U get extra 3h/s, could someone test this further?

### DIFF
--- a/cmake/flags.cmake
+++ b/cmake/flags.cmake
@@ -70,4 +70,12 @@ elseif (CMAKE_CXX_COMPILER_ID MATCHES Clang)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -maes")
     endif()
 
+elseif (CMAKE_CXX_COMPILER_ID MATCHES Intel)
+
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall")
+    set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -O3 -funroll-loops")
+
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -fno-exceptions -fno-rtti")
+    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -O3 -funroll-loops")
+
 endif()
--- a/src/crypto/CryptoNight.h
+++ b/src/crypto/CryptoNight.h
@@ -29,7 +29,7 @@
 #include <stddef.h>
 #include <stdint.h>
 
-#if defined _MSC_VER || defined XMRIG_ARM
+#if defined _MSC_VER || defined XMRIG_ARM || defined __INTEL_COMPILER
 #define ABI_ATTRIBUTE
 #else
 #define ABI_ATTRIBUTE __attribute__((ms_abi))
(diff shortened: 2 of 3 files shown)


Spudz76:
Seems to crash while compiling cn_gpu components when there is not tons of RAM, looks like this:
```
": internal error: ** The compiler has encountered an unexpected problem.
** Segmentation violation signal raised. **
Access violation or stack overflow. Please contact Intel Support for assistance.

icpc: error #10106: Fatal error in /opt/intel/compilers_and_libraries_2019.1.144/linux/bin/intel64/mcpcom, terminated by segmentation violation
compilation aborted for /usr/src/xmrig/src/crypto/cn_gpu_ssse3.cpp (code 1)
```
adding swap does not appear to help, it doesn't seem to use it
```
              total        used        free      shared  buff/cache   available
Mem:          3.8Gi       2.0Gi       219Mi        41Mi       1.6Gi       1.6Gi
Swap:         1.5Gi       2.0Mi       1.5Gi
```

Spudz76:
Slightly different on another machine (even less RAM)
```
": internal error: ** The compiler has encountered an unexpected problem.
** Segmentation violation signal raised. **
Access violation or stack overflow. Please contact Intel Support for assistance.

icpc: error #10105: /opt/intel/compilers_and_libraries_2019.1.144/linux/bin/intel64/mcpcom: core dumped
icpc: warning #10102: unknown signal(-1626568528)
icpc: error #10106: Fatal error in /opt/intel/compilers_and_libraries_2019.1.144/linux/bin/intel64/mcpcom, terminated by unknown
compilation aborted for /usr/src/xmrig/src/crypto/cn_gpu_ssse3.cpp (code 1)
```
the "terminated by unknown" was likely an OOMkill but the other 4GB box did not show a kill just the memory violation
```
              total        used        free      shared  buff/cache   available
Mem:           1.9G        385M        376M        2.7M        1.2G        1.4G
Swap:          2.0G        1.0M        2.0G
```

Spudz76:
skipped that issue by `-DWITH_CN_GPU=OFF` for now
however pico, lite, and heavy are the only algos that run
all CN-2MB variants crash as soon as it gets a job

they do seem slightly faster

`icc version 19.0.1.144 (gcc version 7.4.0 compatibility)`
forced the backing gcc version using `-gcc-name=gcc-7 -gxx-name=g++-7` as when it used gcc-8 (normal Debian Sid system compiler) it crashes hard in some headers - odd since gcc-8 on its own works fine

Spudz76:
Also useragent keeps showing the backing-compiler (gcc) rather than showing icc-19.whatever

Spudz76:
Compilation of that file works with -O1 but not -O2 or -O3
It also crashed on a box with plenty of RAM, such as the first crash.

YetAnotherRussian:
Some details here - https://github.com/xmrig/xmrig/issues/1512#issuecomment-595205254

Please note that both _MSC_VER and __INTEL_COMPILER are difined in case of using Intel Compiler in Visual Studio, so that "ifs" in CryptoNight_x86.h may work not as expected.

cn_gpu_ssse3.cpp crash is a compiler issue I guess, it does not reproduce in old versions.

minhng99:
Stale

Spudz76:
Never really had any beyond-error-margin increase in performance as long as Assembly stubs were being used anyway, it's not like the compiler does anything different to ASM stubs, just all the "glue" code.  Maybe it saved a few nanoseconds between rounds or job dispatch or something leading to perceived/displayed hash increases.  But the pay rate at the pool was never different enough to be worth the extra hassle, and support.

PR #992, opened 2019-03-15
Fix compilation with Clang 3.5

Spudz76:
Fix compilation with Clang 3.5 for those with limited compiler choices
Performance similar to gcc7+ on systems where gcc4 is the only alternative

Based on [official LLVM upstream patch](http://lists.llvm.org/pipermail/cfe-commits/Week-of-Mon-20150518/129343.html) that was merged sometime after 3.5.0


xmrig:
Thank you.

PR #1010, opened 2019-04-05
Add daemon support

hyc:
For solo mining directly against daemon.
Use "daemon+tcp://" URL format instead of "stratum+tcp://"

### DIFF
--- a/src/api/NetworkState.h
+++ b/src/api/NetworkState.h
@@ -50,7 +50,7 @@ public:
 
     char pool[256];
     std::array<uint64_t, 10> topDiff { { } };
-    uint32_t diff;
+    uint64_t diff;
     uint64_t accepted;
     uint64_t failures;
     uint64_t rejected;
--- a/src/base/net/Pool.cpp
+++ b/src/base/net/Pool.cpp
@@ -6,6 +6,7 @@
  * Copyright 2016      Jay D Dee   <jayddee246@gmail.com>
  * Copyright 2017-2019 XMR-Stak    <https://github.com/fireice-uk>, <https://github.com/psychocrypt>
  * Copyright 2018-2019 SChernykh   <https://github.com/SChernykh>
+ * Copyright 2019      Howard Chu  <https://github.com/hyc>
  * Copyright 2016-2019 XMRig       <https://github.com/xmrig>, <support@xmrig.com>
  *
  *   This program is free software: you can redistribute it and/or modify
@@ -65,6 +66,7 @@ xmrig::Pool::Pool() :
     m_enabled(true),
     m_nicehash(false),
     m_tls(false),
+    m_daemon(false),
     m_keepAlive(0),
     m_port(kDefaultPort)
 {
@@ -86,6 +88,7 @@ xmrig::Pool::Pool(const char *url) :
     m_enabled(true),
     m_nicehash(false),
     m_tls(false),
+    m_daemon(false),
     m_keepAlive(0),
     m_port(kDefaultPort)
 {
@@ -135,6 +138,7 @@ xmrig::Pool::Pool(const char *host, uint16_t port, const char *user, const char 
     m_enabled(true),
     m_nicehash(nicehash),
     m_tls(tls),
+    m_daemon(false),
     m_keepAlive(keepAlive),
     m_host(host),
     m_password(password),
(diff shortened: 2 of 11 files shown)


hyc:
I originally wrote this patch against xmrig-proxy, but it seems better to just do it directly in each miner. Presumably the patch against the AMD and Nvidia versions will be about the same as this one. Do you want those submitted separately as well?

xmrig:
I currently busy on miner refactoring, need time to review this PR and make it on my own way. I don't like mixing stratum and HTTP protocol together in one class. In current beta version I added new server side HTTP layer, so need add new client side layer too.
Thank you.

lisergey:
@hyc, why this?

> but it seems better to just do it directly in each miner.

hyc:
@lisergey it was just easier for me to get up and running quickly for testing. For a small number of miners, I think this is preferable, since going through the proxy would add a couple hops worth of latency. Maybe for a very large number of miners, the proxy might have better connection handling than the daemon, but I haven't got a way to test that idea.

Zarkoob:
@hyc I think I have this working. When I load up the miner I get in the log:

`use daemon 10.0.0.10:18081  10.0.0.10.170`
`new job from 10.0.0.10:18081 diff 33621677188 algo cn/r height 1808260`

Does this mean that the miner is "working"?  In monerod when I do a "status" I get a typical status but in there is a "not mining". Is this because the miner is basically using the daemon to get a block difficulty and if it finds a match it sends it to the wallet in the config.json?

reeyon:
It seems to be working. But I can hash with different algo, in this case I use "cn-pico/trtl", and the miner didn't throw error. I'm able to hash higher with this algo against Monero daemon. 
image
The monero diff is so high, it would take ages to accept the job, so I don't know.

hyc:
@Zarkoob correct, the monerod isn't mining, it's just giving the miner a block to work on.

@reeyon the miner doesn't know that it's using the wrong algorithm. You would only find out if you ever found a block and tried to submit it to the daemon; the daemon would reject it then. Basically you're just wasting your power.

Zarkoob:
@hyc Thank you for working on this!  Will the patch eventually work it's way into xmrig-proxy perhaps? I'd like to throw 10-12 rigs at it.

hyc:
@Zarkoob TBH I wasn't sure that patching the proxy would work, due to the 32bit/64bit difficulty discrepancy. But I suppose if the miners are already patched, it could be worth a try.

Zarkoob:
@hyc this would be really awesome for proxy. Would this also work on testnet? I’d hate to have it find a block now and not work right.

hyc:
@Zarkoob All of this was originally tested on testnet. (That's what it's for after all.)

gr8form:
Sorry for the noob question, but couldn't find where else to ask...  Am I better off with a mix of multiple cpu, nvidia & amd gpus at varying hash rate capabilities to use a proxy/private pool; rather than, pointing them all to the same daemon with your modified xmrig?  Bit concerned that I'll be wasting hashes since the difficulty is much higher and might be better off aggregating them via proxy/pool?

Zarkoob:
@hyc ah ok thank you for that. I just got testnet monerod synced and pointed my miner at it and I get this in the daemon:

`[RPC1]	ERROR	cn	src/cryptonote_core/cryptonote_core.cpp:1310	mined block failed verification`

And in the miner I get:

`rejected (1/1) diff 33680 "Block not accepted" (55 ms)`

Is it mining too fast or something?

xmrig:
Replacement for this PR added to [evo](https://github.com/xmrig/xmrig/commits/evo) branch.

Instead of `daemon+tcp://` I added it as `daemon+http://` and `daemon+https://` or dedicated option.

### Examples:
```json
{
	"url": "127.0.0.1:28081",
	"user": "YOUR_WALLET_ADDRESS",
	"pass": "x",
	"rig-id": null,
	"nicehash": false,
	"keepalive": false,
	"variant": "r",
	"enabled": true,
	"tls": false,
	"tls-fingerprint": null,
	"daemon": true,
	"daemon-poll-interval": 1000
}
```

`--url daemon+http://127.0.0.1:28081 -u YOUR_WALLET_ADDRESS -a cn/r`

`--url 127.0.0.1:28081 -u YOUR_WALLET_ADDRESS --daemon --daemon-poll-interval 1000 -a cn/r`.

### Notes
* This will be added into proxy too, but because proxy calculate hashrate only by real accepted shares it will be zero most of time.
* I checked implementation on Monero testnet only, but it should work for other compatible coins too.

hyc:
Please note, there's already an entire previous generation of miners that use "daemon+tcp://" as the URL format. I think it's a mistake to deviate from this now.

lisergey:
I prefer `daemon+https://`
It is really nice that connection between daemon and miners could be over HTTPS, and it should be explicitly noted so.

It is not a mistake to provide new possibilities, even for those who got habit to `daemon+tcp://`, that's just a habit.

xmrig:
Whole url schema part is just compatibly layer, usually don't need explicitly add schema (whatever://) to urls, but I think `daemon+tcp://` is mistake in first place, since it just standard **JSON-RPC over HTTP(s) protocol** and over TCP as well, stratum is JSON-RPC over plain TCP, so `stratum+tcp://` is fully correct.

I don't think I should follow legacy decisions in this case.
Thank you.

xmrig:
v2.15.2-beta released
* https://github.com/xmrig/xmrig/releases/tag/v2.15.2-beta
* https://github.com/xmrig/xmrig-proxy/releases/tag/v2.15.2-beta

islandTraderFX:
The file C:\Users\anons\Desktop\xmrig-master\doc\api\1\config.json is infected with Generic.Application.CoinMiner.1.47B66751 and was moved to quarantine. It is recommended that you run a System Scan to make sure your system is cle

PR #1011, opened 2019-04-11
Config: fix std::max issue with msvc2017

stoffu:
to fix the following errors:

    xmrig\src\core\Config.cpp(165): error C2065: 'max': undeclared identifier
    xmrig\src\core\Config.cpp(165): error C2275: 'size_t': illegal use of this type as an expression


xmrig:
Thank you.

PR #1029, opened 2019-06-08
Return the correct Variant

esfomeado:


xmrig:
It was fine, base for `cn-heavy/tube` is variant 1 with all tweaks was available on algorithm creation date.
Thank you.

PR #1056, opened 2019-07-09
evo branch: Repair colors on Windows (don't test uv_tty_init() return value)

Spudz76:
uv_tty_init() returns negative even on success with Windows
This patch fixes color, tested on Win7 / cmd.exe shell (nothing special)
Does not change anything on Unix etc.

Someone plz test on Win10 and/or PowerShell and whatever other Windows console environments.


Spudz76:
Closed for lack of interest

Issue #1077, opened 2019-07-28
v2.99.2-beta Release Notes


xmrig:
**v2.99.2 is a second major update in 2.99 series, this version introduce NUMA support for RandomX and other algorithms and bug fixes since 2.99.1.**

* **hwloc** dependency added, possible build without hwloc `-DWITH_HWLOC=OFF` but it disable all NUMA related features and new advanced autoconfig.
* If miner built with hwloc RandomX dataset and cache will be allocated on each NUMA node to improve performance.
* New autoconfig use thread affinity, it very important on NUMA hardware.

New options placed in `randomx` object in config file:
```json
{
    ...
    "randomx": {
        "init": -1,
        "numa": true
    },
    ...
}
```
Command line equivalents: `--randomx-init` and `--randomx-no-numa`.

⚠️ Config from other 2.99.x is compatible, but better remove `cpu` object from old config to allow create new threads configuration.

### Previous release notes
* [2.99.0-beta](https://github.com/xmrig/xmrig/issues/1066)

xmrig:
v2.99.2-beta released https://github.com/xmrig/xmrig/releases/tag/v2.99.2-beta if miner still have issues with NUMA please run command `lstopo topo.xml` from hwloc package and open new issue with `topo.xml` and other useful information.
Thank you.

xmrig:
Mining RandomX require 1168 hugepages for dataset and cache per NUMA node + one page per thread, so recommended use `sudo sysctl -w vm.nr_hugepages=1280` (old manuals usually reserve only 128 pages), for 2 NUMA nodes `sudo sysctl -w vm.nr_hugepages=2560` etc.

xmrig:
v2.99.3-beta released https://github.com/xmrig/xmrig/releases/tag/v2.99.3-beta it contains bugfix for AMD FX CPUs.

PR #1087, opened 2019-07-30
Fix compilation against older hwloc (tested Debian Jessie 1.10.0-3)

Spudz76:
Some object types changed define-names, and otherwise seem to work the same.  Leave actual code properly 1.11.x+ compatible while simply copying through the defines when the new names don't exist (assume old hwloc, or crash anyway).

Works well on dual CPU Xeon system.

### DIFF
--- a/src/backend/cpu/platform/HwlocCpuInfo.h
+++ b/src/backend/cpu/platform/HwlocCpuInfo.h
@@ -29,6 +29,15 @@
 #include "backend/cpu/platform/BasicCpuInfo.h"
 
 
+// handle older hwloc where objects had other names
+#ifndef HWLOC_OBJ_PACKAGE
+#   define HWLOC_OBJ_PACKAGE HWLOC_OBJ_SOCKET
+#endif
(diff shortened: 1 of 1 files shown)


Spudz76:
Ref  #1077

xmrig:
Thank you for this PR, but I made it in my way, I use `HWLOC_API_VERSION`, because not nice replace enum values with define names in newer versions. I checked source, all version since 1.11.0 support new names.
Thank you.

Spudz76:
Tested alternate patch, works well, although I added back in the old-version actual version display...

```
 * ABOUT        XMRig/2.99.4-evo clang/9.0.0
 * LIBS         libuv/1.9.0 OpenSSL/1.0.1t hwloc/1.10.0
 * CPU          Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz (2) x64 AES -AVX2
                L2:4.0 MB L3:40.0 MB 16C/32T NUMA:2
```

Spudz76:
Above version string restoration in #1095

PR #1095, opened 2019-08-01
Build hwloc version string based on HWLOC_API_VERSION

Spudz76:
Build hwloc version string based on HWLOC_API_VERSION, whenever `hwlocVersion` object does not exist (<1.11.x)

Probably could just use this method on all versions, as it is more important to display the compiled lib/header version, than whatever the XML tree says its version is (which could be different?).  Maybe complain if they mismatch?


xmrig:
Thank you `HWLOC_API_VERSION` is better than nothing, but it not replacement for version, for example v2.0.4 use API version 2.0.0 and 1.10.1 use 1.10.0.

PR #1146, opened 2019-08-28
Fixed race condition in RandomX thread init

SChernykh:
Thread could deadlock if it started before dataset struct was allocated.
Thanks to crCr62U0 for reporting this.


xmrig:
In addition in branch feature-opencl init sequence changed: mining threads guaranteed starts only after dataset allocated and initialized to allow options like `dataset_host` works without side effects.
Thank you.

PR #1148, opened 2019-08-29
Fix linker marking entire executable as having an executable stack

00-matt:
See: https://wiki.ubuntu.com/SecurityTeam/Roadmap/ExecutableStacks
See: https://wiki.gentoo.org/wiki/Hardened/GNU_stack_quickstart


xmrig:
I will review it little later.
Thank you.

xmrig:
Thank you, I revert changes from one of files since this file used only on Windows https://github.com/xmrig/xmrig/commit/9cfbce5e0992026522beb3608eb79d0f3634568f

00-matt:
Thanks for the quick reply. Will try and package for Gentoo soon.

> Thank you, I revert changes from one of files since this file used only on Windows 9cfbce5

Whoops sorry, shouldn't have added it there. Not familiar with
Windows/PE but I don't think that it needs such a hack.

PR #1191, opened 2019-09-23
Workaround for a bug in binutils-2.32-1 on ARM

SChernykh:
ldr/madd instruction sequence makes compiled binary crash (linker replaces madd with random jump instruction), so separate them.


SChernykh:
Also added a second unrelated bug fix from RandomX repository.

Issue #1235, opened 2019-10-12
device.globalMemSize() returns 0 in ocl_generic_rx_generator.cpp


komatom:
Hey

in file generators/ocl_generic_rx_generator.cpp

line 43, device.globalMemSize() returns 0,

other functions are returning value, for example I tried device.freeMemSize()

Not sure if this is executed too early in the program execution, driver or race issue, same function on listing the available GPUs is working perfectly fine..

I think below calculations are not correct because of that..

I tried to fix it, but for now no luck..

Here is some debug data:

device mem: 0 
dataset_mem: 2315255744
per thread: 2129920
intensity calc: 1472
intensity -= : 512
Device compute units: 32

komatom:
Strange if you call it like below, freeMemSize() before device.globalMemSize() then it works.. I guess OcLib works like that..

```
device.freeMemSize();
const size_t mem = device.globalMemSize();
```

xmrig:
Driver version? anyway it very strange because `device.freeMemSize()` also call `globalMemSize()`, https://github.com/xmrig/xmrig/blob/beta/src/backend/opencl/wrappers/OclDevice.cpp#L160

Might be better if OclDevice cache result of `globalMemSize` to avoid call OpenCL everytime.
Thank you.

komatom:
Hey

Driver version is 19.5.2, but I don't think it matters..

The only difference I see is that freememsize calls oclib from inside the device class, while globalmemsize is called through through a reference/pointer to the device object

Hope you find why... This is seems related to https://github.com/xmrig/xmrig/issues/1199 as it doesn't enable datasethost = true for cards that don't have enough memory

Regards

komatom:
@xmrig I tried to help with that, first time pull request on your project, may need some changes, but it works now with the recent changes in the PR.

Thanks

komatom:
Resolved.

PR #1243, opened 2019-10-14
Fixes OclDevice::globalMemSize() that sometimes returns 0

komatom:
This fixes random return of 0(zero) of globalMemSize() described in this issue https://github.com/xmrig/xmrig/issues/1235


xmrig:
Thank you, looks good, I will merge it tomorrow.

PR #1277, opened 2019-11-12
Fix function names for clang on Apple

SChernykh:


realityworks:
Does this address the libcrypto build issue on Catalina?

PR #1444, opened 2019-12-18
Add Scala (defyx) algorithm [based on MoneroOcean/master]

Spudz76:
Latest code from MoneroOcean/xmrig master, only the Scala `defyx` algorithm part, remastered as properly as I could.  MoneroOcean patchset used more of an add-on method - I just merged it into the main RX files (it is a RandomX variant) so maybe it can be accepted

No supports for backends other than cpu...


xmrig:
Sorry, decision about this algorithm was made a long time ago, I will not support it. I pretty sure @MoneroOcean had own good reasons to support it, but I don't have that reasons.

PR #1434, opened 2019-12-16
RandomSFX (Safex Cash variant) support

SChernykh:
For testing:
```
            "algo": "rx/sfx",
            "coin": null,
            "url": "sfx.pool-pay.com:4025",
            "user": "Safex5zmLUJ1svEg2a4uwrMd2PNCjAXHeP36cyKAv9dd4kKzi8p2mySfFdN6rNyD1Zh4HafPszctQYAykDJkHdhREFh43p1QWtG4T",
```

PR #1445, opened 2019-12-18
Add MoneroV (rx/v) algorithm [based on MoneroOcean/master]

Spudz76:
Latest code from MoneroOcean/xmrig master, only the MoneroV rx/v algorithm part, remastered only slightly (shifted some lines around for consistency)

Appears to also support opencl backend (no CUDA).

### DIFF
--- a/src/backend/opencl/cl/rx/randomx.cl
+++ b/src/backend/opencl/cl/rx/randomx.cl
@@ -8,6 +8,8 @@
 #include "randomx_constants_loki.h"
 #elif (ALGO == ALGO_RX_ARQMA)
 #include "randomx_constants_arqma.h"
+#elif (ALGO == ALGO_RX_V)
+#include "randomx_constants_v.h"
 #endif
 
 #include "aes.cl"
--- a/src/backend/opencl/cl/cn/algorithm.cl
+++ b/src/backend/opencl/cl/cn/algorithm.cl
@@ -21,8 +21,9 @@
 #define ALGO_RX_LOKI        20
 #define ALGO_RX_ARQMA       21
 #define ALGO_RX_SFX         22
-#define ALGO_AR2_CHUKWA     23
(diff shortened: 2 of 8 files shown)


src/backend/opencl/cl/rx/randomx.cl
```diff
@@ -8,6 +8,8 @@
 #include "randomx_constants_loki.h"
 #elif (ALGO == ALGO_RX_ARQMA)
 #include "randomx_constants_arqma.h"
+#elif (ALGO == ALGO_RX_V)
+#include "randomx_constants_v.h"
```
SChernykh: There's no need to add OpenCL constants for rx/v - they're exactly the same as rx/0, see https://github.com/xmrig/xmrig/pull/1434 for reference

SChernykh: Basically, you just need to change algo id in OclRxBaseRunner.cpp from RX_V to RX_0 and that will be enough for OpenCL support. Dataset is generated on CPU with correct Arong2d salt, everything else can just run with rx/0 code.

xmrig:
@Spudz76 Can you explain what is MoneroV is, https://monerov.org/ and https://github.com/monerov/monerov seems already dead, but what is https://monerov.online/ and https://github.com/xmvdev/monerov

SChernykh:
monerov.online is the new "rebooted" MoneroV, it trades on tradeogre instead of the old one now so it's kind of "official".

xmrig:
Oh okay, found https://monerov.online/MoneroV-Swap.pdf

src/backend/opencl/cl/rx/randomx.cl
xmrig: I made required changes.

xmrig:
I was most short lived algorithm ever https://twitter.com/monerovofficial/status/1209509550501482496

PR #1461, opened 2019-12-23
Monero already changed PoW on Nov 30, 2019

suanlian1:
Also minor text corrections.


xmrig:
Packet really called `keepalived` but maybe you right, `keepalive` not looks like a mistake.
Thank you.

PR #1466, opened 2019-12-27
Add CryptoNight Ultra algorithm

### LINKED ISSUE #1399: Add algorithm of Talleo (TLO)
Pull request for xmr-stak: https://github.com/fireice-uk/xmr-stak/pull/2539

This is basically variation of CryptoNight Ultra/Pico with AES mask calculation from original CryptoNight instead of CryptoNight Lite.

This would help bigger rigs mining Talleo to be able to rent their hashing power. Highest total hash rate of known mining pools has been 6.0 MH/s and xmr-stak developers required us to hit 5 MH/s before they add our algorithm.

For testing of implementation:

Official pool: https://pool.talleo.org/
Test address: TA4yACzMYuFYq7V6xVAWYHeS39jQ8w4mKRowpY6NskGuS1rZpjcWuCpdeCypwUCJrK9mGqVW9o1pY2EG3HW7BZkR2YRcc4YNa

mtl1979:
* CryptoNight Ultra is used by Talleo (TLO)

Fixes #1399


xmrig:
I will accept this algorithm only as `cn-pico/tlo`, because it `cn-pico` by memory usage, also it means don't need add extra configuration for this algorithm, there is absolute no reason to generate new configuration if it always will be same with `cn-pico`. This is variant of `cn-pico` not new algorithm, doesn't matter that you use correct mask, turtle was first. 

Other minor requirements:
1. Please create PR as single commit to dev branch, thanks for reverting back donate.h.
2. `test_output_ultra_tlo` not contains full test data.
3. If add this algorithm as variant of `cn-pico` don't need to change CnAlgo, correct mask will be calculated by generic functions.

mtl1979:
We tried with using algorithm name "cn-pico/tlo" but some functions rely on base algorithm to decide what parameters to use, so the code would not work without changing parts that might break support for Ultra Lite and we can only test that mining Talleo works...

I tested that ```test_output_ultra_tlo``` contains what the code expects it to contain, last bytes were never used, so I zeroed them out instead of guessing the contents.

xmrig:
About zero bytes, try this configuration `[ [5, -1] ],` also if self test passed it means implementation is correct.
Thank you.

mtl1979:
Replaced the zeroed data with output from [5, -1] as CPU config... Gonna rebase and squash after everything works... Gonna monitor now for rejected shares... Got them when donation was enabled, that's why I test with donation set as 0%.

When I tried before renaming to cn-ultra, I got hard crash on Cuda, so something goes wrong somewhere if algorithm is set as cn-pico/tlo instead of cn-ultra/tlo... I can't explain why it happens, might be some specialization is not selected correctly. Same happens if I disable standard CryptoNight in thread configuration as some parts of the code are shared.

xmrig:
For xmrig-cuda you should also use dev branch, recently added alternative API, algorithm passed as string, old API where algorithm passed as number deprecated and no longer used.

As alternative you can show `test_output_ultra_tlo` and I do renaming changes byself, at least you trying submit it to upstream very hard, unlike many others. 
Thank you.

mtl1979:
I only fixed so self-test works, because the developer who made the original patch couldn't get it to work correctly... For me, it is more important to get simple support added instead of trying to work with untested experimental code.

xmrig:
Okay I will add this algorithm.

xmrig:
Replaced by https://github.com/xmrig/xmrig/commit/402c44b547820c9c4fc9a2a66183c387483f983e

PR #1496, opened 2020-01-12
zlib needed on some macos

jtgrassie:
Needed for (at least MacPorts based) macOs builds.


jtgrassie:
Latest master doesn't seem to need. Closing.

PR #1502, opened 2020-01-15
Optimizations for AMD Bulldozer

SChernykh:
- Added support for XOP instructions
- Enabled Ryzen code for Bulldozer because it's faster there too


UnclWish:
Tried dev version with buldozer optimizations...
loki algo
xmrig 5.5.1 - 1 core from FX 8320 - 350-385 h/s
xmrig dev - 1 core from FX 8320 - 250-285 h/s

SChernykh:
Can you try to set `"wrmsr":false` in config and compare again? Also please double check that xmrig allocates 100% huge pages every time and nothing else is running during hashrate tests.

UnclWish:
With "wrmsr":false and restarting miner speed became about 350+ h/s - normal.
Without restarting with reloading config on the fly nothing changes. MSR reloads only with full miner restart.
And with "wrmsr":true speed 250 h/s.

SChernykh:
It seems that MSR values that worked for Piledriver don't work for Bulldozer (I tested it on Opteron 6338P). Too bad, I hoped it would be the same.

UnclWish:
Maybe MSR values for Buldozer are differs?

UnclWish:
Sorry for offtop: you're russian?

SChernykh:
Yes.

UnclWish:
Can we speak on russian language here? )))

Lonnegan:
May be the optimization works for the original bulldozer (e.g. FX-8150), but with piledriver (e.g. FX-8350) version 5.1.1 is faster, 1330 H/s vs. 1300 H/s. Not much difference, but reproducible.

SChernykh:
@Lonnegan Can you try to set "asm" to false in config.json and test 5.5.3 again? This will turn off AMD-specific optimizations.

SChernykh:
@Lonnegan I've been told that someone else with AMD FX-8320E (which is also Piledriver) got better hashrate on 5.5.3 - it went up from 2300 to 2313 h/s. So maybe there is something interfering with mining on your PC.

Lonnegan:
I'll try tonight.

Lonnegan:
Ok, it was my fault! I compared 5.5.1-msvc vs. 5.5.3-gcc by accident. 5.5.3-msvc is indeed about 20 H/s faster even on Piledriver based CPU! :) Good job!

PR #1529, opened 2020-02-02
Crash fix for Bullodzer CPUs

SChernykh:


micron10:
hear have problem with latest 5.5.3vand bulldozer
machine is crash and reboot

SChernykh:
@micron10 It's better to create an issue here: https://github.com/xmrig/xmrig/issues and add full miner output and miner config there.

PR #1553, opened 2020-02-15
RandomXHFPI

klaroooo:
Creating pull request to include RandomXHFPI for Klaro Electronic Money.

More info on Klaro: https://bitcointalk.org/index.php?topic=5223800.0

"Klaro is inspired by cutting edge technology and it's purpose, in a phase where the world seems in "danger" we decided to bring you Klarity, Klaro is electronic money, Klaro is a tool which allows the creation of what "mankind" call money in an electronic way. Klaro shouldn't be seen as crypto watchers, investors, traders see cryptocurrencies, rather than source of income/fun/mining or learning, cryptocurrencies should be seen as precious code left by several developers around the world which has real value ( you can also see it as artifacts). Today we introduce you Klaro, an electronic form of money with nothing more than our honest vision of the future as "mankid" evolves and each individual tries to map it's future ahead."...

### DIFF
--- a/src/crypto/randomx/bytecode_machine.hpp
+++ b/src/crypto/randomx/bytecode_machine.hpp
@@ -52,6 +52,7 @@ namespace randomx {
 			const int_reg_t* isrc;
 			const rx_vec_f128* fsrc;
 		};
+        const int_reg_t* isrc2;
 		union {
 			uint64_t imm;
 			int64_t simm;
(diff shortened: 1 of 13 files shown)


src/crypto/randomx/bytecode_machine.hpp
```diff
@@ -205,6 +206,14 @@ namespace randomx {
 			*ibc.fdst = rx_mul_vec_f128(*ibc.fdst, *ibc.fsrc);
 		}
 
+		static void exe_FMUL2I_R(RANDOMX_EXE_ARGS) {
+			 const rx_vec_f128 a = _mm_cvtepi32_pd(_mm_set_epi64x(0, *ibc.isrc)); 
+			*ibc.fdst = rx_mul_vec_f128(*ibc.fdst, a);
+			 const rx_vec_f128 b = _mm_cvtepi32_pd(_mm_set_epi64x(0, *ibc.isrc2));
+			*ibc.fdst = rx_mul_vec_f128(*ibc.fdst, b);
+			*ibc.fdst = rx_sqrt_vec_f128(*ibc.fdst);
```
SChernykh: Either of `a` and `b` can be negative resulting in `sqrt(negative value)` here. This won't work reliably as a PoW.

PR #1558, opened 2020-02-20
not working after built

newtothis280:
built
added pool to config - didn't work
then added as m_pools.push_back(new Url("pool.example.com", 3333, "WALLET", "PASSWORD")); in xmrig-silent/src/common/config/CommonConfig.cpp line 63
checked with process explorer and not working


newtothis280:
also added the wallet and pool in base64 where required to do so - still not working

Spudz76:
You should ask whoever makes "modrig" which I have never heard of

this is a support site for xmrig

PR #1592, opened 2020-03-10
AVX2 optimized code for AstroBWT

SChernykh:
Added "astrobwt-avx2" parameter in config.json, it's turned off ("false") by default.

4-5% speedup on CPUs with proper AVX2 support (AMD Ryzen starting with Zen2, Intel Core starting with Haswell).

There will be no speedup on the following CPUs:

- Intel Pentium/Celeron don't support AVX2
- AMD Zen/Zen+ have only half-speed AVX

GCC compiled version is faster without AVX2, MSVC compiled version is faster with AVX2

### DIFF
--- a/cmake/astrobwt.cmake
+++ b/cmake/astrobwt.cmake
@@ -23,6 +23,16 @@ if (WITH_ASTROBWT)
             src/crypto/astrobwt/salsa20_ref/salsa20.c
         )
     else()
+        if (CMAKE_SIZEOF_VOID_P EQUAL 8)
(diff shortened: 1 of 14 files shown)


Lonnegan:
AMD Ryzen 7 2700 with Windows 10 x64 v1909, xmrig 5.10.0 msvc, 16T with cpu binding:

AVX2 off = 655 H/s
AVX2 on = 753 H/s

So even with Zen/Zen+ AVX2 is an advantage

PR #1675, opened 2020-05-13
Fix Windows include files cases

ybh1998:
When cross-compiling on case sensitive systems, such as Linux, there will be errors.
Change cases of Windows include files and link library.


xmrig:
Would be nice if you also share a manual on how to do a cross compile.
Thank you.

PR #1718, opened 2020-06-07
Stop linker from making stack executable

00-matt:
Same as #1148.


xmrig:
Merged into dev and evo branch.
Thank you.

PR #1771, opened 2020-07-10
Adopt new SSE2NEON and reduce ARM-specific changes

jserv:
This patch updated SSE2NEON [1], which contains more functions
provided by Intel intrinsics, only implemented with NEON-based
counterparts to produce the exact semantics of the intrinsics.
Consequently, ARM-specific changes against CryptoNight_arm can
be reduced as well.

[1] https://github.com/DLTcollab/sse2neon/


xmrig:
Thank you.

PR #1774, opened 2020-07-13
RandomX: added cache QoS support

SChernykh:
False by default. If set to true, all non-mining CPU cores will not have access to L3 cache.


2010phenix:
and what we have with this L3 cache optimize for miner and for Win OS lag?

PR #1818, opened 2020-08-26
Fix OpenSSL static link when BUILD_STATIC=ON is provided

antoris123:


xmrig:
#1640 Already merged into dev branch.
Thank you.

PR #1823, opened 2020-09-04
RandomX: added parameter for scratchpad prefetch mode

SChernykh:
`scratchpad_prefetch_mode` can have 4 values:

0: off
1: use `prefetcht0` instruction (default, same as previous XMRig versions)
2: use `prefetchnta` instruction (faster on Coffee Lake and a few other CPUs)
3: use `mov` instruction

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -21,7 +21,8 @@
         "rdmsr": true,
         "wrmsr": true,
         "cache_qos": false,
-        "numa": true
+        "numa": true,
+        "scratchpad_prefetch_mode": 1
     },
     "cpu": {
         "enabled": true,
--- a/src/core/config/Config_default.h
+++ b/src/core/config/Config_default.h
@@ -51,7 +51,12 @@ R"===(
     "randomx": {
         "init": -1,
         "mode": "auto",
-        "numa": true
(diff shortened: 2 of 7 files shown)


Lonnegan:
Can you please explane for non-programmers what the different parameters do and where the advantages and disadvantages might be? Is there an advantage on systems, that are not dedicated for mining but are used for daily work when prefetch is set to 0, so that inner core traffic is reduced? Why is prefetchnta faster on Coffee Lake than prefetcht0? Might mov be an advantage on CPUs, that have bad built-in prefetchers and if, what CPUs are that for example? :)

SChernykh:
I don't have answers to some of these questions. I only observed that mode 2 is a tiny bit faster on some Intel CPUs, but didn't find a clear pattern. There should be no difference for other programs running together with miner.

kowalabearhugs:
I observe a small hashrate increase on some Intel Xeon Westmere CPUs when switching prefetch modes.

2x xeon X5650 w/ 24gb in tri-channel (6x 4gb sticks)
msr: enabled
huge pages: true
1gb pages: true

prefetch mode 1
~3250 h/s

prefetch mode 2
~3320 h/s

PR #1831, opened 2020-09-13
reserve at most 1 bit for wrapping detection

cohcho:
Add Job::nonceMask() to know mutable bits
Reserve at most 1 bit within Nonce::next to detect counter wrapping


xmrig:
Thank you.

PR #1834, opened 2020-09-15
soft_aes: fix previous optimization

cohcho:
It was faster to inline aesenc, aesdec without any modifications during previous optimization #1163


xmrig:
Replaced by #1835
Thank you.

PR #1865, opened 2020-10-05
Update README.md

DeadManWalkingTO:
Update README.md

### DIFF
--- a/src/version.h
+++ b/src/version.h
@@ -22,7 +22,7 @@
 #define APP_ID        "xmrig"
 #define APP_NAME      "XMRig"
 #define APP_DESC      "XMRig miner"
-#define APP_VERSION   "6.16.4-dev"
+#define APP_VERSION   "6.16.4"
 #define APP_DOMAIN    "xmrig.com"
 #define APP_SITE      "www.xmrig.com"
 #define APP_COPYRIGHT "Copyright (C) 2016-2021 xmrig.com"


xmrig:
Please explain this PR and others similar to this.
Thank you.

DeadManWalkingTO:
Add/Fix Hyperlinks to project(s).
Replacing Absolute URLs with Relative URLs.
Add Table of Contents.
Replacing Hyperlinks of last Release and last Release date to latest release .

heliosgnosis:
not to mention for any new miners most of the "windows" batch syntax's commands are incorrect,  someone new will waste half a day just because of an extra - or double for old school syntax use.  This has been an issue for months actually,  and in irony friends I turned to mining had this issue and indeed wasted hours when all it was well as so listed above,  this probably has cost you all massive amounts of dev revenue without your knowing,   I have used this program on windows and linux since the genesis's day's but now we have a massive trove of newbies,   might want to fix that pretty QUICK,  other than that thank you for all your hard work after all this time but yeah... all of us old timers know better and how to change if program does not load on windows or linux but the influx to which will only increase crowd not so much,   but it is your revenue not mine,  I am just letting you know it is broken,  has not been updated and the will work with so and so version + is not true whatsoever at least not on windows platform the main platform of focus for newbies who would not know a simple change of 1 - is the only issue.  Cheers

Issue #1871, opened 2020-10-06
Error in randomx.cpp:463 in version 6.3.4 and 6.3.5


BillGatesIII:
**Describe the bug**
xmrig: /home/user/xmrig/src/crypto/randomx/randomx.cpp:463: randomx_vm* randomx_create_vm(randomx_flags, randomx_cache*, randomx_dataset*, uint8_t*, uint32_t): Assertion `cache == nullptr || cache->isInitialized()' failed.

**To Reproduce**
Start xmrig 6.3.4 or 6.3.5.

**Expected behavior**
No error, like in 6.3.3.

**Required data**
 * ABOUT        XMRig/6.3.5 gcc/8.3.0
 * LIBS         libuv/1.38.1 OpenSSL/1.1.1g hwloc/2.2.0
 * HUGE PAGES   supported
 * 1GB PAGES    supported
 * CPU          Intel(R) Xeon(R) CPU E5-2620 v2 @ 2.10GHz (2) x64 AES
                L2:3.0 MB L3:30.0 MB 12C/24T NUMA:2
 * MEMORY       6.7/62.8 GB (11%)
 * DONATE       1%
 * ASSEMBLY     auto:intel
 * POOL #1      xmrpool.eu:5555 algo cn/r
 * COMMANDS     hashrate, pause, resume, results, connection
 * OPENCL       disabled
 * CUDA         disabled
[2020-10-06 18:02:52.504]  net      use pool xmrpool.eu:5555  54.37.7.208
[2020-10-06 18:02:52.504]  net      new job from xmrpool.eu:5555 diff 50000 algo rx/0 height 2202466
[2020-10-06 18:02:52.504]  cpu      use argon2 implementation SSSE3
[2020-10-06 18:02:52.514]  msr      register values for "custom" preset has been set successfully (10 ms)
[2020-10-06 18:02:52.514]  randomx  init datasets algo rx/0 (24 threads) seed aeb044aa31eab166...
[2020-10-06 18:02:54.066]  randomx  #0 allocated 3072 MB huge pages 100% (1553 ms)
[2020-10-06 18:02:54.066]  randomx  #1 allocated 3072 MB huge pages 100% (1553 ms)
[2020-10-06 18:02:54.066]  randomx  -- allocated 6144 MB huge pages 100% 6/6 (1553 ms)
[2020-10-06 18:03:04.282]  randomx  #1 dataset ready (10215 ms)
[...]

PR #1872, opened 2020-10-06
RxVm: fix randomx_create_vm call

cohcho:
randomx_create_vm requires either cache or dataset, but not both


cohcho:
Two and more instances of RxDataset with 1 GiB pages will have at least one allocated but not initialized cache.
That's a source of problem in #1871

SChernykh (approved):

PR #1873, opened 2020-10-07
RandomX: allocate 2 MB pages for generated code, if possible

SChernykh:
+0.2% boost on Ryzen 7 3700X


SChernykh:
Hashrate is very unstable, it needs more testing.

PR #1880, opened 2020-10-08
RandomX: remove duplicate SuperscalarHash generation

syjcnss:
SuperscalarHash generates twice in initCacheCompile. Fix it.


SChernykh (changes requested):
Your changes break dataset generation completely, I'm getting 100% rejected shares now. Please test your modification on the real Monero pool first. And as far as I can see, this code is called only once for dataset generation, there is no duplication.

syjcnss:
I misunderstood the code.

PR #1883, opened 2020-10-09
Hashrate: display numerical error

cohcho:
Different workers store stats asynchronously and with different frequency.
It has direct impact on accuracy of hashrate calculations.
Error calculation overhead balanced with faster iteration due to tail index

- Hashrate is being calculated as previously
- Error is being displayed when it's above zero

### DIFF
--- a/src/backend/common/Hashrate.cpp
+++ b/src/backend/common/Hashrate.cpp
@@ -34,12 +34,18 @@
 #include "base/io/json/Json.h"
 #include "base/tools/Chrono.h"
 #include "base/tools/Handle.h"
+#include "base/io/log/Log.h"
 
 
-inline static const char *format(double h, char *buf, size_t size)
+inline static const char *format(double h, char *buf, size_t size, double errorDown, double errorUp)
 {
     if (std::isnormal(h)) {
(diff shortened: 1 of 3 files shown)


cohcho:
https://github.com/cohcho/xmrig/commit/cd23d2ea3af9ccc0a67ad02430adcce1c08839ef

Joking optimization to simulate hashrate jump due to error but not actual speed up.
Naturally error jumps around shares submit.

SChernykh:
Compilation error in MSVS 2019:
```
Hashrate.cpp(186,26): error C2664: 'xmrig::Hashrate::Value::Value(xmrig::Hashrate::Value &&)': cannot convert argument 1 from 'rapidjson::Type' to 'const xmrig::Hashrate::Value &'
```
I fixed it temporarily by changing the line to `rapidjson::Value out(rapidjson::kArrayType);`. I'll keep testing.

SChernykh:
The displayed error values are way overestimated and can only confuse users. I think it's just better to refactor hashrate calculation to take the maximal common interval among all threads where they're all mining and display hashrate from that interval or `n/a` if that interval is too short.

PR #1889, opened 2020-10-12
uv: fix performance issue

cohcho:
unix implementation of uv_async_t has been wasting cpu cycles for nothing since 1.29.0 release
implement efficient callback scheduling for linux

### DIFF
--- a/src/base/io/Async.cpp
+++ b/src/base/io/Async.cpp
@@ -1,0 +1,78 @@
+#include "base/io/Async.h"
+
+
+#if defined(XMRIG_UV_PERFORMANCE_BUG)
+#include <sys/eventfd.h>
+#include <sys/poll.h>
+#include <unistd.h>
+#include <cstdlib>
+
+
+namespace xmrig {
+
+
+uv_async_t::~uv_async_t()
+{
+    close(m_fd);
+}
+
+
+static void on_schedule(uv_poll_t *handle, int status, int events)
+{
+    static uint64_t val;
+    uv_async_t *async = reinterpret_cast<uv_async_t *>(handle);
+    for (;;) {
+        int r = read(async->m_fd, &val, sizeof(val));
+
+        if (r == sizeof(val))
+            continue;
+
+        if (r != -1)
+            break;
+
+        if (errno == EAGAIN || errno == EWOULDBLOCK)
+            break;
+
+        if (errno == EINTR)
+            continue;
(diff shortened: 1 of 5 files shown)


cohcho:
https://github.com/libuv/libuv/pull/2231
The source of performance drop and it's likely will not be fixed soon.

xmrig:
Looks good, but need some time for review.
Thank you.

src/base/io/Async.cpp
```diff
@@ -0,0 +1,78 @@
...
+{
+    static uint64_t val;
+    uv_async_t *async = reinterpret_cast<uv_async_t *>(handle);
+    for (;;) {
+        int r = read(async->m_fd, &val, sizeof(val));
+
+        if (r == sizeof(val))
+            continue;
+
+        if (r != -1)
```
SChernykh, line 30: Comparing uint64_t variable with -1 is sketchy. Should it be int64_t instead?

src/base/io/Async.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+
+int uv_async_send(uv_async_t *async)
+{
+    static const uint64_t val = 1;
+    int r;
+    do {
+        r = write(async->m_fd, &val, sizeof(val));
+    }
+    while (r == -1 && errno == EINTR);
```
SChernykh, line 68: Same here, uint64_t is compared with -1.

src/base/io/Async.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+
+static void on_schedule(uv_poll_t *handle, int status, int events)
+{
+    static uint64_t val;
+    uv_async_t *async = reinterpret_cast<uv_async_t *>(handle);
+    for (;;) {
+        int r = read(async->m_fd, &val, sizeof(val));
+
+        if (r == sizeof(val))
```
SChernykh, line 27: I don't understand this part. If it successfully reads 8 bytes (uint64_t), it just keeps reading it again. How is async called then?

SChernykh (changes requested):

src/base/io/Async.cpp
SChernykh: Ok, I figured it out. It reads all incoming data and then fails with EAGAIN, triggering async call.

SChernykh (approved):

PR #1894, opened 2020-10-14
Cache Hash support

En4orcer:
Cryptonight cache hash is a modified variation of cryptonight conceal.

https://letshash.it/cxche/ - current working pool supporting cache hash


En4orcer:
http://explorer.cxche.org/index.html <- We've just passed 19k testnet blocks with this algorithm

PR #1912, opened 2020-10-23
MSR: supress kernel module warning

cohcho:
https://github.com/torvalds/linux/commit/a7e1f67ed29f0c339e2aa7483d13b085127566ab#diff-2f768554951616110618f5c2f6b3a580b570ca63a9bbbe6b3300adf66ee84443
Writes can still be allowed by supplying the `allow_writes=on` module parameter


cohcho:
Linux: fix writes to MSR

Direct MSR writes will be filtered since Linux >5.8.0
Since there is no alternative to direct MSR access yet then add parameter
  `allow_writes=on` to supress harmless warning

PR #1948, opened 2020-11-15
Add GitHub Actions compiling

Randomblock1:
This will automatically compile XMRig on Linux, MacOS, and Windows every time a commit is made and produce a (statically linked) build artifact by using GitHub Actions. Also adds a shiny new badge.
It also does this to pull requests, in addition to normal pushes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,4 +100,4 @@ jobs:
         uses: actions/upload-artifact@v2.2.1
         with:
           name: xmrig-windows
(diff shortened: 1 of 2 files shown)


xmrig:
We already use for it own self hosted [buildbot](https://buildbot.xmrig.com/#/) for it, it also compiles every commit and files created by buildbot used as releases.
Thank you.

Randomblock1:
Ah cool! You should link that somewhere in the readme, in case someone wants to download a dev build without compiling it. I didn’t even know it existed.

PR #1949, opened 2020-11-15
custom DNS support for xmrig

mr-459:
hi
sometimes dns serevrs get problem with our proxy servers
please add ability to use multiple custom dns's to resolve pool address(if 1 dns got problem , other dns maybe still working)
thanks a lot


wule108:
maybe add ability to use multiple custom dns's to resolve pool address , like  --dns=8.8.8.8 .   And The pool domain name record cannot be updated in time through the system default dns server

PR #1979, opened 2020-12-15
Don't use team 'x64' for '64-bit'

Low-power:
`x64` is short for `x86-64`, which only refers to this specific architecture made by AMD; using the term for any other 64-bit architectures such as `arm64` is wrong. Please simply use `64-bit` to indicate a 64-bit architecture.

Source file `src/backend/cpu/platform/BasicCpuInfo_arm.cpp` would still output JSON that uses `x64` to indicate 64-bit support, should it be fixed as well?

### DIFF
--- a/src/Summary.cpp
+++ b/src/Summary.cpp
@@ -88,11 +88,11 @@ static void print_cpu(Config *)
 {
     const auto info = Cpu::info();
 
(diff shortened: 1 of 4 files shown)


xmrig:
Merged into dev branch.
Thank you.

PR #1980, opened 2020-12-15
Fix build on GCC 11

voidanix:
Currently fails with `error: 'size_t' was not declared in this scope;`, should everything be moved to `std::size_t`?


xmrig:
Merged into dev branch.
Thank you.

PR #1986, opened 2020-12-19
Dataset initialization with AVX2 (faster startup)

SChernykh:
20-30% faster dataset initialization on some CPUs:
- Zen3 (all)
- Zen2 without SMT (some of mobile Ryzen 4000 CPUs)
- Intel CPUs without HT (some Core i3 and i5)

New config option `init-avx2` controls it:

-1 = Auto detect
0 = Disabled (use old unoptimized code)
1 = Always enabled (only on CPUs that support AVX2)


Lonnegan:
Just for better understanding: the advantage is exclusively at the start of the software, not during runtime, correct?

SChernykh:
Faster start and faster dataset update every 2048 blocks (~3 days).

PR #2004, opened 2020-12-24
Update config.json

adrianus678:


agentpatience:
I am experiencing a strange issue when I start Monero 17.1.7 daemon and successfully sync my node with block.txt.

When I mine on the same computer my node dumps with “Killed” message. I set my xmrig to thread count 34 instead of 36 for 2695v4 server but it still does not remain stable.

PR #2039, opened 2021-01-15
Fixed solo mining

SChernykh:
It was broken since 6.7.0


xmrig:
[v6.7.2](https://github.com/xmrig/xmrig/releases/tag/v6.7.2) released.
Thank you.

PR #2055, opened 2021-01-22
Add missing "cstdio" library

GoDzM4TT3O:
Compilation fails if the above library is missing. This fixes a compilation error.

### DIFF
--- a/src/hw/dmi/DmiReader_unix.cpp
+++ b/src/hw/dmi/DmiReader_unix.cpp
@@ -29,7 +29,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
-
+#include <cstdio>
 
 #ifdef __FreeBSD__
 #   include <kenv.h>


xmrig:
1. On what operating system is the error?
2. Compiler version.
3. What error exactly.

PR itself is OK, but curious about the error conditions, since it successfully builds on all buildbot targets.
Thank you.

GoDzM4TT3O:
>     1. On what operating system is the error?
> 
>     2. Compiler version.
> 
>     3. What error exactly.
> 
> 
> PR itself is OK, but curious about the error conditions, since it successfully builds on all buildbot targets.
> Thank you.

1. Arch Linux; Linux Kernel 5.10.9
2. GCC 10.2.0
3. If I recall correctly, it was an error explicitly saying that this library was missing, and I remember that functions like `fputs` were highlighted in red (meaning undefined)

xmrig:
Thank you.

GoDzM4TT3O:
No problem!

PR #2086, opened 2021-02-06
Does the source code support the cpu of armv7 system?Dev

xlmshack:
Does the source code support the cpu of armv7 system?


SChernykh:
No, only armv8 (64-bit) is supported.

PR #2089, opened 2021-02-07
Optimized cn-heavy for Zen3

SChernykh:
- Uses scratchpad interleaving to access only the closest L3 slice from each CPU core.
- Also activates MSR mod for cn-heavy because CPU prefetchers get confused with interleaving
- 7-8% speedup on Zen3

### DIFF
--- a/src/backend/cpu/CpuBackend.cpp
+++ b/src/backend/cpu/CpuBackend.cpp
@@ -81,6 +81,7 @@ public:
 
     inline void start(const std::vector<CpuLaunchData> &threads, size_t memory)
     {
+        m_workersMemory.clear();
         m_hugePages.reset();
         m_memory    = memory;
         m_started   = 0;
@@ -95,8 +96,10 @@ public:
         if (ready) {
             m_started++;
 
-            m_hugePages += worker->memory()->hugePages();
-            m_ways      += worker->intensity();
+            if (m_workersMemory.insert(worker->memory()).second) {
+                m_hugePages += worker->memory()->hugePages();
+            }
+            m_ways += worker->intensity();
         }
         else {
             m_errors++;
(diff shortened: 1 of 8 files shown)


ScardracS (approved):
I also have a Zen3 architecture and these modifies work as expected

SChernykh:
Binaries for testing: https://download.xmrig.com/xmrig/6.8.2-dev/91151ce4a1a14fe9e75f6a99bd3a77a55540379c/xmrig-6.8.2-dev-msvc-win64.zip (login `xmrig` password `download`).

ScardracS:
> Binaries for testing: https://download.xmrig.com/xmrig/6.8.2-dev/91151ce4a1a14fe9e75f6a99bd3a77a55540379c/xmrig-6.8.2-dev-msvc-win64.zip (login `xmrig` password `download`).
> 
> Delete thread config for `cn-heavy` from config.json before running because correct thread count and affinity is important for this to work properly!

I have moved from 1.1kH/s to a solid 1.6kH/s. Good job!

Lonnegan:
Effectively no change with a Ryzen 5 5600X. 720 H/s (Heavy/XHV) with 6.8.1 and 725 H/s with 6.8.2.

SChernykh:
Did you regenerate thread config for 5600X? It went from 755 to 815 h/s on my 5600X.

Lonnegan:
[1, 0],
            [1, 2],
            [1, 4],
            [1, 6],
            [1, 8],
            [1, 10],
            [1, 9],
            [1, 11]

Which CN-Heavy coin have you tested? Perhaps there are differences between the variants. I've tested Haven (XHV)

SChernykh:
I tested XHV, 5600X @ 4.65 GHz, same thread affinity

Lonnegan:
You've tested with locked core clock, I've tested with default turbo. Perhaps the new optimization utilizes the ressources more intense so the turbo can't boost as high as before.

SChernykh:
Fixed clock is the only way to get reproducible and comparable results, also fixed clock + lower voltage better for mining because default settings use too high voltage.

PR #2095, opened 2021-02-11
Update xmrig.cpp

ScardracS:


xmrig:
I usually update the year when the content of the file is changed.
Thank you.

PR #2104, opened 2021-02-14
Added `pause-on-active` option

SChernykh:
Windows only for now. When set to true, pauses mining when user touches mouse or keyboard.

### DIFF
--- a/src/base/kernel/Platform.h
+++ b/src/base/kernel/Platform.h
@@ -49,6 +49,7 @@ public:
     static inline const String &userAgent() { return m_userAgent; }
 
     static bool isOnBatteryPower();
+    static bool isUserActive();
 
 private:
     static char *createUserAgent();
--- a/src/base/kernel/Platform_unix.cpp
+++ b/src/base/kernel/Platform_unix.cpp
@@ -158,3 +158,10 @@ bool xmrig::Platform::isOnBatteryPower()
     }
     return false;
 }
+
+
+bool xmrig::Platform::isUserActive()
+{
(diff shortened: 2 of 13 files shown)


Lonnegan:
This feature ist so cool for mining systems, which are used for daily work, as well. I've put in 5 as a parameter for pause-on-active, so xmrig stops mining when I touch the keyboard or the mouse and resumes 5 seconds after the last input. So you can mine with full power, even with normal priority without being without slowed down by xmrig.

jondwillis:
The documentation should probably be updated to note that this is windows-only until it is ported to Mac + Linux.

xmrig:
I added note about supported OS https://xmrig.com/docs/miner/config/misc#pause-on-active macOS is supported, but Linux support likely will be never added, Linux is mostly headless.
Thank you.

Lonnegan:
Ok, is it possible that the function only works if xmrig is running in the same account as the user handling the mouse and keyboard? I have the problem on a PC that xmrig continues to run despite the function being activated when another logged-in user moves the mouse and keyboard.

SChernykh:
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getlastinputinfo
> However, GetLastInputInfo does not provide system-wide user input information across all running sessions. Rather, GetLastInputInfo provides session-specific user input information for only the session that invoked the function.

This is done for security reasons.

Lonnegan:
Hm, that's bad, but thank you 4 info :)

PR #2130, opened 2021-02-24
pause on active option

taherkaraki:


SChernykh:
Thanks for report, we'll forward your botnet version to antivirus companies ASAP!

PR #2141, opened 2021-02-28
Correctly handle submit block response for self select

sdbondi:
The self-select client handles a response from `submitblock` requests.

Previously, the submit block responses would result in an error message
and enter Retry state because parseResponse expects all responses to be
for `getblocktemplate`.

Removed an unused `m_results` map field that would continue growing in size 
as blocks are submitted.


hansieodendaal:
Looking good! I Will revert when I have run some tests.

Issue #2098, opened 2021-02-12
msr: Write to unrecognized MSR 0xc0011020 by xmrig (AMD Ryzen 5 - 4600H)


Bogdan107:
This is message in my syslog:
```msr: Write to unrecognized MSR 0xc0011020 by xmrig (pid: 68425). Please report to x86@kernel.org.```

$ uname -p
```AMD Ryzen 5 4600H with Radeon Graphics```
$ uname -r
```5.10.15-gentoo-x86_64```

Some messages from xmrig:
```
* ABOUT        XMRig/6.8.1 gcc/10.2.0
* LIBS         libuv/1.40.0 OpenSSL/1.1.1i hwloc/2.4.0
* CPU          AMD Ryzen 5 4600H with Radeon Graphics (1) 64-bit AES
               L2:3.0 MB L3:8.0 MB 6C/12T NUMA:1
* MOTHERBOARD  HP - 87B1
* ASSEMBLY     auto:ryzen
cpu      use argon2 implementation AVX2
msr      register values for "ryzen_17h" preset have been set successfully (21 ms)
 ```

May be this report will help.

coolhaircut:
The linux kernel is disabling userspace writes to MSR in version 5.9 :(

For now, you should be able to fix this by adding `msr.allow_writes=on` to the kernel cmdline (`GRUB_CMDLINE_LINUX` variable in `/etc/default/grub')

PR #2161, opened 2021-03-05
Added Userspace MSR permissions clarification in CPU.md

coolhaircut:
Userspace MSR writes are no longer allowed by default in Linux Kernels starting with 5.9. _(Per this [commit](https://git.kernel.org/pub/scm/linux/kernel/git/tip/tip.git/commit/?h=x86/misc&id=a7e1f67ed29f0c339e2aa7483d13b085127566ab) and [explaination](https://www.phoronix.com/scan.php?page=news_item&px=Linux-Filter-Tightening-MSRs).)_

Userspace MSR writes in kernels that are locked down by default can still be enabled if the `msr.allow_writes=on` kernel flag is set. I.e. setting  `GRUB_CMDLINE_LINUX_DEFAULT="msr.allow_writes=on"` in `/etc/default/grub` for most linux distros.

### DIFF
--- a/doc/CPU.md
+++ b/doc/CPU.md
@@ -91,11 +91,11 @@ RandomX mining mode: `auto`, `fast` (2 GB memory), `light` (256 MB memory).
 #### `1gb-pages`
 Use 1GB hugepages for RandomX dataset (Linux only). Enabled (`true`) or disabled (`false`). It gives 1-3% speedup.
 
+#### `wrmsr`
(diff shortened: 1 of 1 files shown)


coolhaircut:
Relevant to issue #2098 and will be a common question as this is started to become the default in some distributions (Ubuntu 21.04 specifically.)

DeeDeeRanged:
Debian on which Ubuntu is based has the the issue too. Specially Debian 11 (bullseye/testing)

PR #2194, opened 2021-03-19
fix keepalive config from being ignored

jtgrassie:
When a pool reports it supports keepalive, but a client has set
keepalive:false or keepalive:0 in the config, this was ignored and
keepalive was always enabled.


jtgrassie:
Fixes #2195

jtgrassie:
Current behavior is how @xmrig intends, pool enforced keepalive, hence closing this PR.

PR #2217, opened 2021-03-29
setBlob should run after setAlgorithm

esrrhs:
Hi, is there a bug? because the function xmrig::Job::setBlob call nonceOffset and nonceSize which use the algorithm family, so if algorithm family change after setBlob there may be a bug

### DIFF
--- a/src/base/net/stratum/Client.cpp
+++ b/src/base/net/stratum/Client.cpp
@@ -377,6 +377,14 @@ bool xmrig::Client::parseJob(const rapidjson::Value &params, int *code)
         return false;
     }
 
+    const char *algo = Json::getString(params, "algo");
+    if (algo) {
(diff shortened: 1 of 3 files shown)


SChernykh:
Technically you are correct, but if you check the code it only matters for KawPow which uses EthStratumClient where it's already in the correct order. So there's no bug for other algorithms here.

esrrhs:
> 
> 
> Technically you are correct, but if you check the code it only matters for KawPow which uses EthStratumClient where it's already in the correct order. So there's no bug for other algorithms here.

yes, but maybe it could be more robustness if change the call order

SChernykh (approved):

xmrig:
Merged into dev branch.
Thank you.

PR #2220, opened 2021-03-31
Implement pause-on-active for Linux and FreeBSD

In-line:


SChernykh:
It's not a good idea to bring X dependencies, making compilation harder just for this small insignificant feature. In many cases xmrig runs on a server without X installed.

In-line:
@SChernykh We can make it optional feature.

In-line:
I can change the code to load necessary libraries with `dlopen`

xmrig:
This feature can be added only if it does not require X11 to build the miner.
Thank you.

In-line:
@xmrig I submitted second version #2222 with build-time optional X11 and runtime optional X11 libs.

So we can build with X11 headers present and then ship it for servers & gui desktops as single executable.

PR #2223, opened 2021-04-01
Add LTO/IPO support in CMake

In-line:
| LTO | Results (sudo xmrig --bench=1M) 30% completion |   
| - | -|
| Yes | 10s/60s/15m 2006.0 1982.9 n/a H/s max 2058.7 H/s 
| No | 10s/60s/15m 1915.0 1867.4 n/a H/s max 2030.7 H/s |

```
 * ABOUT        XMRig/6.10.0 gcc/10.2.0
 * LIBS         libuv/1.41.0 OpenSSL/1.1.1k hwloc/2.4.0
 * HUGE PAGES   supported
 * 1GB PAGES    disabled
 * CPU          Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz (1) 64-bit AES
                L2:1.0 MB L3:8.0 MB 4C/4T NUMA:1
 * MEMORY       13.4/31.2 GB (43%)
                DIMM_A0: 16 GB DDR4 @ 2133 MHz CMK32GX4M2B3000C15  
                ChannelA-DIMM1: <empty>
                DIMM_B0: 16 GB DDR4 @ 2133 MHz CT16G4DFD824A.C16FBD
                ChannelB-DIMM1: <empty>
 * MOTHERBOARD  Gigabyte Technology Co., Ltd. - B150M-D3H-CF
```

https://stackoverflow.com/questions/23736507/is-there-a-reason-why-not-to-use-link-time-optimization-lto
https://gcc.gnu.org/onlinedocs/gccint/LTO-Overview.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,6 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.9)
+cmake_policy(SET CMP0069 NEW)
+
 project(xmrig)
 
 option(WITH_HWLOC           "Enable hwloc support" ON)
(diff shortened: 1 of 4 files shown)


00-matt:
Why not just leave the builder to set `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON` instead of them setting `-DWITH_IPO=ON` and running through all this extra code?

SChernykh:
In my tests there was no improvement for RandomX at all, also this adds dependency on cmake 3.9. It's better to leave it to the command line as @00-matt suggests.

In-line:
@00-matt Fair point

Issue #2224, opened 2021-04-01
Build with -DWITH_OPENCL=OFF fails


00-matt:
##### Describe the bug

XMRig fails to build when it is configured with `-DWITH_OPENCL=OFF`.

```
[75/208] Building CXX object CMakeFiles/xmrig.dir/src/backend/cuda/CudaBackend.cpp.o
FAILED: CMakeFiles/xmrig.dir/src/backend/cuda/CudaBackend.cpp.o 
[...]

PR #2225, opened 2021-04-01
Add missing include

00-matt:
`memory` header ends up not being included when built without OpenCL support.

Closes: https://github.com/xmrig/xmrig/issues/2224


PaulinaParangerHr:
.patch
Duplicate of #

PR #2273, opened 2021-04-16
Mac M1 build update proposal.

devnexen:
This supports the crc32 word size various flavors instructions, also
 fixing profiling build for arm64.

### DIFF
--- a/src/crypto/rx/Profiler.h
+++ b/src/crypto/rx/Profiler.h
@@ -50,9 +50,15 @@ static FORCE_INLINE uint64_t ReadTSC()
 #ifdef _MSC_VER
     return __rdtsc();
 #else
+#if defined(__x86_64__) || defined(__amd64__) || defined(__i386__)
     uint32_t hi, lo;
     __asm__ __volatile__("rdtsc" : "=a"(lo), "=d"(hi));
     return (((uint64_t)hi) << 32) | lo;
+#elif defined(__aarch64__)
+    uint64_t cur;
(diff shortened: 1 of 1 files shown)


SChernykh:
It's better to PR to https://github.com/DLTcollab/sse2neon because new sse2neon.h will be taken from there eventually.

cmake/cpu.cmake
```diff
@@ -28,13 +28,13 @@ if (ARM_TARGET AND ARM_TARGET GREATER 6)
         set(XMRIG_ARMv8 ON)
         add_definitions(/DXMRIG_ARMv8)
 
-        CHECK_CXX_COMPILER_FLAG(-march=armv8-a+crypto XMRIG_ARM_CRYPTO)
+        CHECK_CXX_COMPILER_FLAG(-march=armv8-a+crypto+crc XMRIG_ARM_CRYPTO)
```
SChernykh, line 31: None of the supported algorithms use crc32 so there's no point in adding it to `-march`. Also, the way this change is done, it will fall back to `armv8-a+crc` if `armv8-a+crypto+crc` flag is not available - this will disable AES instructions even if they are available.

devnexen: ok may do specific change for the Mac M1 then

SChernykh: But there is no point in adding this at all. XMRig doesn't use CRC32.

devnexen: ah you re right

PR #2283, opened 2021-04-19
support builtin msr

jsonboss:
Some custom linux kernel may be builtin msr, please support such case. Thanks.


Spudz76:
Seems like this should work okay in all use-cases.

xmrig:
Merged into dev branch with changes https://github.com/xmrig/xmrig/commit/e26fbc96e9a343bcf6d9b9d1d7a121ce2b5089d9
Thank you.

PR #2347, opened 2021-05-06
XMRig as service

denisw2009:
i suggest to add new feature in XMRig. Could you add to XMRig ability to work as Windows service. It will be fine if XMRig can start with windows in background and will not disturb user with --pause-on-active function


ShaddyR:
So, work_as_service for windows was realized?
Or can you modify pause-on-active mechanism or add other variant for multiuser system like server RDP? To watch if any session is active and pause waiting while all are deactivated? As variant - something like "green mode" - minimalistic cpu threads (configurable) and priority (cpu time) using if there are any active user sessions.

PR #2379, opened 2021-05-16
Fix CL code for KawPow where it assumes everything is AMD

Spudz76:
This simple change allows Apple M1 OpenCL to actually work (1.51MH/s)

Probably other non-AMD platforms also (Intel iGPUs?)

On actual device with `cl_amd_media_ops` it should operate identically as before.  On those without it will no longer crash realtime miner kernel compilation.


xmrig:
Generated files wasn't include final `#endif` I fixed it by regenerating files https://github.com/xmrig/xmrig/commit/08a2c143f5a65912ee8e5b50ddb69f937ac4d1da.
Thank you.

Spudz76:
Yep, missed that after hotfixing, thanks.

PR #2394, opened 2021-05-20
Convert ARM_TARGET option() to selection-list

Spudz76:
CMake `option()` shortcut for cached entries is intended only of ON/OFF (boolean) items.

Convert to string selection-list which works smooth in ccmake/cmake-gui

May reduce having to force with `-DARM_TARGET=8` to override the boolean nature of the option when possibly re-coerced to ON or OFF during config or generation steps.  Did not observe such a bug, but there are definitely some sort of bugs around that option working as intended.


Spudz76:
Fixed up logic, tested on ARMv8 debian VM.

PR #2432, opened 2021-06-09
Fix shellcheck warnings in enable_1gb_pages.sh

candrews:
Checked using www.shellcheck.net

Specific issues addessed:
* https://github.com/koalaman/shellcheck/wiki/SC2046
* https://github.com/koalaman/shellcheck/wiki/SC2044

### DIFF
--- a/scripts/enable_1gb_pages.sh
+++ b/scripts/enable_1gb_pages.sh
@@ -1,12 +1,9 @@
-#!/bin/bash -e
+#!/bin/sh -e
 
 # https://xmrig.com/docs/miner/hugepages#onegb-huge-pages
 
-sysctl -w vm.nr_hugepages=$(nproc)
+sysctl -w vm.nr_hugepages="$(nproc)"
 
-for i in $(find /sys/devices/system/node/node* -maxdepth 0 -type d);
-do
-    echo 3 > "$i/hugepages/hugepages-1048576kB/nr_hugepages";
-done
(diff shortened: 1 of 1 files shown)


Valenzuelafer (commented):
No way

Valenzuelafer:
Akin to

scripts/enable_1gb_pages.sh
```diff
@@ -1,12 +1,9 @@
-#!/bin/bash -e
```
captain-hoity-toity, line 1: The shell script is generally well-written and follows good practices. However, here are a few recommendations to improve it:

```markdown
1. **Add a Description:** It's a good practice to add a description at the top of the script explaining what it does. This will help others to understand the script's purpose.

2. **Error Handling:** The script uses the `-e` flag to exit on error, which is good. However, it might be beneficial to add more error handling, for instance, checking if the `sysctl` and `find` commands succeed.

3. **Check for Root Permissions:** This script appears to require root permissions to run successfully (due to the `sysctl` and `echo` commands). It would be useful to add a check at the start of the script to ensure it's being run with the necessary permissions.

4. **Use More Descriptive Variable Names:** The `{}` placeholder in the `find` command is not very descriptive. Consider using a more descriptive placeholder.

5. **Quote Your Variables:** It's good practice to quote your variables in shell scripts to prevent word splitting and pathname expansion.
```

PR #2445, opened 2021-06-16
Support for solo mining with miner signatures (Wownero)

SChernykh:
CPU only for now because elliptic curve operations are required for mining.

To test:
- Compile https://git.wownero.com/wownero/wownero/src/branch/dev-v0.10
- start `wownerod --testnet --offline --fixed-difficulty 10000`
- wownerod --start-mining ADDRESS --spendkey SPENDKEY
- wait until it mines past block height 60
- `stop_mining` in wownerod
- run `xmrig -o 127.0.0.1:11181 --coin wownero -u ADDRESS --spend-secret-key SPENDKEY --daemon`


peme14k:
#2445SChernykh:miner_signature

kl1059:
I suggest adding 
--spend-secret-key-from-file filename
its always been bad practice to put passwords etc in the command line argument on linux because of the 'ps axw' command, image accounting, or system crash dumps that contain all the commands running with their args.  Someone could upload a kernel crash dump to their favorite distro support team and disclose their spendkey (which i am guessing is bad).

admittedly most linux systems are single user systems now so the risk is low

SChernykh:
There's a `config.json` file for this, command line parameters are just duplicates of what's available there.

Valenzuelafer:
Need help

Valenzuelafer (commented):
Monero.mine

PR #2454, opened 2021-06-25
Dev Always deleted by antivirus software, please release a version that will not be recognized as a virus.

wxysystem32:
Always deleted by antivirus software, please release a version that will not be recognized as a virus.
image


Spudz76:
Impossible, whitelist the folder.

PR #2484, opened 2021-07-15
ZeroMQ support for solo mining

SChernykh:
Gets new blocks from daemon immediately without polling, saving ~0.5 seconds on average when daemon gets new block from the network. Also saves some CPU cycles because it doesn't need to poll daemon every second.

Testing: add "daemon-zmq-port": 28083 to xmrig's pool config in config.json and run ./monerod --testnet --zmq-pub tcp://127.0.0.1:28083


Deep0310:
Help me set github

Pada tanggal Kam, 15 Jul 2021 16.28, SChernykh ***@***.***>
menulis:

PR #2512, opened 2021-08-07
Show the number of transactions in pool job

SChernykh:
Useful to check if pool/proxy is working properly and can also be used to compare different pools.


SChernykh:
But it's always visible

Lonnegan:
Yes, you are right. Why couldn't I see it yesterday?! Downloaded the wrong version? Forget my last post, please. ;-)

PR #2548, opened 2021-08-19
Added automatic coin detection for daemon mining

xmrig:
Previously added wallet address parser now used to detect coin by user specified wallet address for daemon mining. That mean `coin` and `algo` option is not required for following coins:
- Monero
- Wownero
- Sumokoin 
- ArQmA
- Wownero
- DERO

List of coins may extend in future.


dodgeyan11223:
how to download software and start mining?

PR #2583, opened 2021-09-16
Fix --threads generates "*" profile without "kawpow":false to negate it.

Spudz76:
When generating a catch-all Asterisk thread profile for CPU backend, it enables kawpow.

Fix is to generate a `kawpow:false` to negate the catch-all and ensure kawpow is always deeply disabled for CPU.

### DIFF
--- a/src/core/config/ConfigTransform.cpp
+++ b/src/core/config/ConfigTransform.cpp
@@ -43,6 +43,7 @@ static const char *kAsterisk    = "*";
 static const char *kEnabled     = "enabled";
 static const char *kIntensity   = "intensity";
 static const char *kThreads     = "threads";
+static const char *kKawPow      = "kawpow";
 
 
 static inline uint64_t intensity(uint64_t av)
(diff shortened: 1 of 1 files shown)


xmrig:
1. KawPow always disabled for CPU https://github.com/xmrig/xmrig/blob/dev/src/backend/cpu/CpuConfig.cpp#L112 or I missed something?
2. Don't need define "kawpow" string it already exists as `Algorithm::kKAWPOW_RVN`.

Thank you.

Spudz76:
1. When patched with the `algo-perf` feature KawPow does attempt to act as if it's viable, and fails to init KawPow on CPU (obviously unworkable) and then sits there forever.  This is because `algo-perf` will test any algo that has a configuration, whether the backend actually supports it or not, and the asterisk profile makes every algo have a configuration.  This patch forces KawPow to always be disabled even when using the threads option.  There is no reason to not set it false when generating a catch-all profile for CPU.  IMO there is no reason to even use `--threads` but people keep doing it anyway.

2. Algorithm.h is not included in the scope of that file as far as I could tell and importing it for one string seemed dumb.

Just trying to have less junk in the MO patch set that could be atomically upstreamed and reduce maintenance when rebasing (restoring cn-gpu and adding panthera and algo-perf is plenty of non-upstreamable work to have to deal with).  For example and IMO, `algo-perf` should be included in core since at least three pools support it (adhoc new stratum extension) but it's currently a bit hacky how it integrates and was there before the internal benchmarking feature.  So I also understand why it's not really clean enough to accept as-is.  It seems like it should use the Benchmark class now that it exists, too, but it might not know how to test anything but RandomX.

xmrig:
> importing it for one string seemed dumb.

With importing the resulting binary will contain only one `kawpow` string, without 2.
Thank you.

Issue #2574, opened 2021-09-07
Unable to find entry point GetHostNameW in WS2_32.dll


jututt:
When trying launching latest 6.15.0 on my Win7 x64 machines, both GCC and MSVC, it crashes and displays error message as per title (roughly translated to english by me).
Previous 6.14.1 works fine

xmrig:
Seems libuv dropped support for Windows 7 https://github.com/libuv/libuv/issues/3260 
Thank you.

Spudz76:
Thanks for the heads-up, I will be running Win7 until the end of ESU ([January 10, 2023](https://docs.microsoft.com/en-us/lifecycle/faq/extended-security-updates)) guess the workaround is to not upgrade libuv unless they reverse their terrible decision.

SChernykh:
#2586

jututt:
closed with 6.15.1

PR #2586, opened 2021-09-17
Fixed Windows 7 compatibility

### LINKED ISSUE #2585: No working
When I compile doesn't work because GetHostNameW is missing from ws2_32.dll. I'm using window 7.

SChernykh:
Fixes #2585


xmrig:
Also fixes #2574

PR #2599, opened 2021-09-22
Improve compiler version strings

Spudz76:
* Add full MSVC version info
* Add ICC version display including backing GCC version (previously just said backing GCC version)


xmrig:
I will not merge this PR, reasons:
 
- I don't like the verbose version for MSVC, just 2019 is good enough for the purpose.
- Intel compiler support is outdated, it doesn't handle the new "one" compiler based on Clang.
- Recent version of the classic Intel compiler screwed up `__INTEL_COMPILER`, it just 2021 number now, which is not a full version.
- Slowly upcoming base handles all versions in single place with unified access https://github.com/xmrig/base/blob/master/kernel/Versions.cpp.

Thank you.

PR #2600, opened 2021-09-22
possible race fix in Worker

eaxeax:
The variable m_count is being read and written concurrently, in theory it is necessary to fix this

### DIFF
--- a/src/backend/common/Worker.h
+++ b/src/backend/common/Worker.h
@@ -20,6 +20,9 @@
 #define XMRIG_WORKER_H
 
 
+#include <atomic>
+
+
 #include "backend/common/interfaces/IWorker.h"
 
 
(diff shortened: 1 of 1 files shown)


SChernykh:
Not needed. Each worker runs in a separate thread, so there's only 1 thread writing to each `m_count` instance. Reading in parallel with writing will return either old or new value which is ok for hash counting (next value will be read on the next iteration).

eaxeax:
This is UB, maybe then it's better to use memory_order_relaxed for store and load?

SChernykh:
There's a good rule in programming `don't fix what's not broken`. GPU worker also writes and reads m_count from different threads by the way. I explained already in my previous message what "UB" really means in this case, it works perfectly fine on both x86 and ARM. The only place that maybe needs fixing is `CpuWorker<N>::hashrateData` because it reads `m_count` twice and has 0.0000...0001% chance of reading different values. But even if it happens, it'll be fixed on the very next iteration, and it's only used to display hashrate. Atomic is not needed here. It's the main mining loop and it's better not to add unnecessary overhead.

PR #2621, opened 2021-10-10
Add pause-on-process feature

Spudz76:
Was requested on Reddit.  I thought it would be ridiculous and too complex but it turned out to be pretty easy.

Tested on Win7 and Linux (I have no Win10 anywhere to test).  No MacOS support yet, but stubbed out so it shouldn't crash.

May be able to roll back some sections to c++11 but everything works apparently the same with c++17 anyway.

If `WITH_PAUSE_PROCESS` is off the resulting code is identical to without this patch (everything in ifdef wrapping).  Including no added libs and c++11 compatible.

Configuration is such as `"pause-on-process": ["/usr/bin/vlc"],` you can list many and it's substring case insensitive of the entire commandline on Linux including arguments, or the fully qualified EXE path on Windows (no arguments).

### DIFF
--- a/src/base/kernel/Platform.h
+++ b/src/base/kernel/Platform.h
@@ -67,6 +67,8 @@ private:
     static char *createUserAgent();
 
     static String m_userAgent;
+    static uint8_t m_processListTicks;
+    static bool m_processListState;
 };
 
 
--- a/src/base/kernel/Platform_unix.cpp
+++ b/src/base/kernel/Platform_unix.cpp
@@ -193,8 +193,21 @@ bool IsNumeric(const std::string& s)
     return !s.empty() && std::all_of(s.begin(), s.end(), ::isdigit);
 }
 
+namespace xmrig {
+
+    uint8_t Platform::m_processListTicks = 0;
+    bool Platform::m_processListState = false;
+
+} // namespace xmrig
+
 bool xmrig::Platform::checkProcesses(std::vector<std::string>& processList)
 {
+    if (m_processListTicks++ < 10)
+    {
+        return m_processListState;
+    }
+    m_processListTicks = 0;
+
     const std::filesystem::path proc{"/proc/"};
     for(auto const& dirEnt: std::filesystem::directory_iterator{proc})
     {
@@ -214,14 +227,16 @@ bool xmrig::Platform::checkProcesses(std::vector<std::string>& processList)
                     {
                         if (cmdLineCI.find(processName.c_str()) != std::string::npos)
                         {
-                            return true;
+                            m_processListState = true;
+                            return m_processListState;
                         }
                     }
                 }
             }
         }
     }
-    return false;
+    m_processListState = false;
+    return m_processListState;
 }
 #endif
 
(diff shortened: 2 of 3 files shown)


SChernykh:
I'm strongly against it. The whole process enumeration will eat CPU on every timer tick (2 times a second), also opening every single process will trigger all antiviruses that are not triggered yet (even if this code is not called, antiviruses will see these API calls in the binary).

P.S. The code itself leaks `wszProcessName` on every iteration.

Spudz76:
Weird, it works fine on my Windows 7. I was watching the supported OS levels on things...

Agree on the leaks I expected some cleanups.

Essentially what I told the people asking for the feature but they wanted it pretty bad.  Maybe it could be on a ~15s schedule rather than where I stuck it next to the other `pause-on-` features to mitigate polling.  Also it doesn't poll at all if the configuration is empty so it would only "hurt" people who take the extra steps to configure it.

But yes I hate it too.  I wish the OS themselves had an API for "is <string> running?" to avoid all the iteration, but I couldn't find any.  I didn't like having to use old dirent stuff therefore the leap forward to c++17 specifically for the std::filesystem stuff.  Probably could tag that as a property just for Platform_unix.c and keep the global at c++11.  For less possible impact / compiler compatibility with ones I don't test (gcc8/9/10, clang14, msvc latest).  I tried to keep it somewhat global language compatible but it probably chokes on wack file paths anyway.  Plenty of strcasestr implementations don't do i18n at all which would definitely only work correctly for ASCII, thus the pulling shlwapi for its Locale-aware case-insensitive strstr implementation.  The linux char-class hack probably isn't so great at anything but ASCII / untested.

Could also try using a method that subscribes to an event based notification, whatever Music Players or Virus Scanners etc use so the game launch becomes an event to go into chill-mode, instead of having to poll every process constantly.  But I didn't think most of those were standard (maybe Linux with dbus? but who wants to implement that, seems like it must be worse from bloat/dep standpoint even though it behaves better).

SChernykh:
https://docs.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-enumprocessmodulesex
https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-queryfullprocessimagenamea
Both say minimum supported client Windows Vista
Re: antivirus - they can reclassify it from CoinMiner to something worse which is not ideal.

Spudz76:
And Vista was before 7 though, so all good.

True I suppose higher "threat level" might put it in a class where it's impossible to whitelist.  But my logic was it already needs to be whitelisted so functionally who cares.  Although there is no other way to obtain the process names and wouldn't other task-manager tools end up virally flagged?  I think it also works using the slightly lesser permission of `PROCESS_QUERY_LIMITED_INFORMATION` which is supported in Vista+ also and probably still provides the exename (but not more sensitive process handle access) which should make it more Not A Virus friendly behavior.

Will make some mods.

Spudz76:
Fixed up the string leaks and eliminated some garbage once I figured out you can use `std::wstring` and the `&wstr[0]` notation directly with the windows functions.  Then the `std::wstring` automatic cleanup and bounds checking applies itself.

It does work fine with the limited permissions, and didn't need `PROCESS_VM_READ` at all (bad example copied).

PR #2636, opened 2021-10-18
AstroBWT speedup (up to +7%)

SChernykh:
Speedup depends on how saturated memory bandwidth is.

CPU | threads | h/s before | h/s after | speedup
-|-|-|-|-
Core i7-2600 | 8 | 267.8 | 268.9 | +0.4%
Ryzen 7 4700U | 8 | 447.5 | 457.7 | +2.3%
Ryzen 5 5600X @ 4.65 GHz, DDR4-3800 CL14 | 12 | 1636.9 | 1746.8 | +6.7%

### DIFF
--- a/src/crypto/astrobwt/AstroBWT.cpp
+++ b/src/crypto/astrobwt/AstroBWT.cpp
@@ -86,15 +86,54 @@ static void Salsa20_XORKeyStream_AVX256(const void* key, void* output, size_t si
 }
 #endif
 
-void sort_indices(int N, const uint8_t* v, uint64_t* indices, uint64_t* tmp_indices)
+static inline bool smaller(const uint8_t* v, uint64_t a, uint64_t b)
+{
+	const uint64_t value_a = a >> 21;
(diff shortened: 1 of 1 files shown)


Spudz76:
This is a useless report, but I see a definite 2.63% gain on Celeron J3455 (Apollo Lake), which I found surprising. `117.61 -> 120.7` H/s

Possibly similar gain for old AMD Phenom II X6 1035T but very hard to tell it only scores ~36H/s so a few percent would be fractional (and it has a normal ripple).  Also DDR2 lololol.

Confirm probably the 0.4% on Sandy Bridge DDR3 stuff, similar to the old AMD it's hard to tell if just normal +-1.2H/s ripple.

Spudz76:
Yes I reconsidered testing that way for less variance / repeatability.  Most of these also are driving GPUs with PhoenixMiner although  that is "somewhat" light on CPU.  Bad test cases all around (but there is an average gain I can see on pool already from a few hours running).

PR #2740, opened 2021-11-28
Added VAES support for Cryptonight variants

SChernykh:
VAES instructions are available on Intel Ice Lake/AMD Zen3 and newer CPUs.
+4% GhostRider speedup on Ryzen 5 5600X.


Spudz76:
Broke compile with < gcc-10

Previous gcc's `avxintrin.h` does not contain `_mm256_loadu2_m128i` or `_mm256_storeu2_m128i`

Clang appears to have it all the way back to at least 7.  MSVC all good as well...

Spudz76:
Fixed for me #2746

PR #2743, opened 2021-11-28
Fix minor leaks and some unintialized components (valgrind)

Spudz76:
Also fix benchmark feature when `WITH_RANDOMX=OFF`


src/crypto/cn/CryptoNight_x86.h
```diff
@@ -679,7 +679,7 @@ inline void cryptonight_single_hash(const uint8_t *__restrict__ input, size_t si
     }
 
     for (size_t i = 0; i < props.iterations(); i++) {
-        __m128i cx;
+        __m128i cx = _mm_setzero_si128();
```
SChernykh, line 682: No. Don't add stuff to the main loop because some program says so. It was carefully crafted to run as fast as possible.

src/crypto/cn/CryptoNight_x86.h
```diff
@@ -1116,7 +1116,8 @@ inline void cryptonight_double_hash(const uint8_t *__restrict__ input, size_t si
     uint64_t idx1 = al1;
 
     for (size_t i = 0; i < props.iterations(); i++) {
-        __m128i cx0, cx1;
+        __m128i cx0 = _mm_setzero_si128();
+        __m128i cx1 = _mm_setzero_si128();
```
SChernykh, line 1120: Same, don't add stuff to the main loop.

src/crypto/cn/CryptoNight_x86.h
```diff
@@ -1375,7 +1376,10 @@ void cryptonight_quad_hash_zen(const uint8_t* __restrict__ input, size_t size, u
     uint64_t idx2 = al2;
     uint64_t idx3 = al3;
 
-    __m128i cx0, cx1, cx2, cx3;
+    __m128i cx0 = _mm_setzero_si128();
+    __m128i cx1 = _mm_setzero_si128();
+    __m128i cx2 = _mm_setzero_si128();
+    __m128i cx3 = _mm_setzero_si128();
```
SChernykh, line 1382: Same. Even though it's not the main loop, it's not needed. Probably a false positive.

SChernykh (changes requested):

Spudz76:
Those only showed up during the GR optimizer phase and were quite spammy.  I hadn't seen them before with normal CN algos, somehow.  Had to patch them to see the other leaks, will revert those mainloop inits.

Spudz76:
I think zeroing the buf and hash within ghostrider actually fixed the uninitialized condition in the CN code which is how it doesn't show up except in the GR optimizer phase.  I also patched that after patching the CN code so that makes sense.

Spudz76:
Adjusted whitespace since I found a code-standard precedent for using `{ 0 }` and not `{0}` :)

PR #2769, opened 2021-12-01
Performance fixes

SChernykh:
- Fixed several performance bottlenecks introduced in v6.16.1
- Fixed overall GCC-compiled build performance, it's the same speed as MSVC build now
- **Linux builds are up to 10% faster now compared to v6.16.0 GCC build**
- **Windows builds are up to 5% faster now compared to v6.16.0 MSVC build**


ahorek:
xmrig --bench=250K -a gr --rotation 15

Where `rotation` is an integer between 0 and 19 (inclusive).

mr459:
i just started mining with v 6.16.0 and 6.16.2
they had the same hashrate
i found no 5% add or etc

SChernykh:
@mr459 6.16.2 is faster than 6.16.0. Did you compare hashrate on the same rotation or 24 hour average hashrate? What OS and what xmrig builds (GCC or MSVC) did you compare?

PR #2789, opened 2021-12-04
Adding /d switch

Anime4000:
1. Adding `/d` to change current drive in addition to changing current directory for a drive.
2. Enclose quote `"%~dp0"` allow space in `cd` command (to be safe).
3. Allow user run this script without set **Administrator** in properties.

Running `*.cmd` As Administrator will spawn on `C:\WINDOWS\system32` and break if `xmrig.exe` not in `C:\` (Default UAC)


matheusbach:
Hi. Alternatively you can put your settings in the config.json file and just open the .exe executable as admin

matheusbach (approved):

PR #2793, opened 2021-12-05
CN: Consistency cleanup

Spudz76:
* Remove `props.isHeavy()` logic within the `_gr_` functions as they are no longer called unless already `CN_GR_[0-5]`
* Add `IS_CN_HEAVY_XHV` to match existing `IS_CN_HEAVY_TUBE` and use it everywhere the long-form test was used
* Rename `IS_HEAVY` to `IS_CN_HEAVY` to better match the `IS_CN_HEAVY_(TUBE|XHV)`
* Use `IS_CN_HEAVY` the same everywhere `props.isHeavy()` was called separately
* Ensure `IS_CN_HEAVY` is also defined everywhere `CN_STEP4` macro was used, as its `props.isHeavy()` was changed
* Flip ordering of `cn_vaes_enabled` tests for slightly improved readability, and allows to not validate the other bools unless VAES is available (now in roughly likelihood-to-be-false and importance order)

Tested compile, and valid accepts for `cn-heavy/xhv` and `ghostrider` but I do not have any VAES capable rigs, so it still needs a diligence proof test on something that does have it.  Also don't have any pools to test the `cn-heavy/0` or `cn-heavy/tube` but I think the `cn-heavy/xhv` test proves those by association.

Should be a tiny bit nicer since all these are validated once into constexpr-bool's the same way everywhere.


src/crypto/cn/CryptoNight_x86.h
```diff
@@ -293,9 +293,10 @@ template<Algorithm::Id ALGO, bool SOFT_AES, int interleave>
 static NOINLINE void cn_explode_scratchpad(cryptonight_ctx *ctx)
 {
     constexpr CnAlgo<ALGO> props;
+    constexpr bool IS_CN_HEAVY = props.isHeavy();
 
 #   ifdef XMRIG_VAES
-    if (!SOFT_AES && !props.isHeavy() && cn_vaes_enabled) {
+    if (cn_vaes_enabled && !SOFT_AES && !IS_CN_HEAVY) {
```
SChernykh, line 299: cn_vaes_enabled (non compile-time check) must be the last in list, or compiler won't be able to remove the whole `if` block when one of the other two compile-time checks returns false.

src/crypto/cn/CryptoNight_x86.h
```diff
@@ -407,15 +408,15 @@ template<Algorithm::Id ALGO, bool SOFT_AES, int interleave>
 static NOINLINE void cn_implode_scratchpad(cryptonight_ctx *ctx)
 {
     constexpr CnAlgo<ALGO> props;
+    constexpr bool IS_CN_HEAVY = props.isHeavy();
 
 #   ifdef XMRIG_VAES
-    if (!SOFT_AES && !props.isHeavy() && cn_vaes_enabled) {
+    if (cn_vaes_enabled && !SOFT_AES && !IS_CN_HEAVY) {
```
SChernykh, line 414: cn_vaes_enabled must be last

src/crypto/cn/CryptoNight_x86.h
```diff
@@ -1068,7 +1073,7 @@ inline void cryptonight_double_hash_asm(const uint8_t *__restrict__ input, size_
     }
 
 #   ifdef XMRIG_VAES
-    if (!props.isHeavy() && cn_vaes_enabled) {
+    if (cn_vaes_enabled && !IS_CN_HEAVY) {
```
SChernykh, line 1076: cn_vaes_enabled must be last

src/crypto/cn/CryptoNight_x86.h
```diff
@@ -1116,7 +1121,7 @@ inline void cryptonight_double_hash_asm(const uint8_t *__restrict__ input, size_
...
```
SChernykh, line 1124: cn_vaes_enabled must be last

src/crypto/cn/CryptoNight_x86.h
```diff
@@ -1302,7 +1310,7 @@ inline void cryptonight_double_hash(const uint8_t *__restrict__ input, size_t si
     }
 
 #   ifdef XMRIG_VAES
-    if (!SOFT_AES && !props.isHeavy() && cn_vaes_enabled) {
+    if (cn_vaes_enabled && !SOFT_AES && !IS_CN_HEAVY) {
```
SChernykh, line 1313: cn_vaes_enabled must be last

src/crypto/cn/CryptoNight_x86.h
```diff
@@ -1506,7 +1514,7 @@ inline void cryptonight_double_hash(const uint8_t *__restrict__ input, size_t si
...
```
SChernykh, line 1517: cn_vaes_enabled must be last

src/crypto/cn/CryptoNight_x86.h
```diff
@@ -1842,7 +1852,7 @@ inline void cryptonight_quad_hash(const uint8_t *__restrict__ input, size_t size
...
```
SChernykh, line 1855: cn_vaes_enabled must be last

src/crypto/cn/CryptoNight_x86.h
```diff
@@ -1905,7 +1915,7 @@ inline void cryptonight_quad_hash(const uint8_t *__restrict__ input, size_t size
...
```
SChernykh, line 1918: cn_vaes_enabled must be last

SChernykh (changes requested):

SChernykh:
> Flip ordering of cn_vaes_enabled tests for slightly improved readability, and allows to not validate the other bools unless VAES is available (now in roughly likelihood-to-be-false and importance order)

Do you know how compile-time checks are optimized by the compiler? They are essentially zero-cost.

SChernykh:
Same goes for all `pros.isHeavy()` replacements - these are already compile time and your change doesn't change anything in the generated x64 code.

Spudz76:
OK, following those rules, I've gone the other direction then.

Consistency is when everything is done the same way, whichever way that is.  The latest commit pushes everything back the other direction instead (use `props.isWhatever()` for everything, drop all unneeded local constexprs shadowing the props).

eiaeea:
@Spudz76 Hi Tony, what more is needed to add rx/yada support to Xmrig??

PR #2815, opened 2021-12-15
Fixed cn-heavy in 32-bit builds

SChernykh:


hammadmallick:
is there a way i can run XMrig on a 32 bit windows?

Issue #2876, opened 2022-01-19
Failure to compile on armv7l: ‘vld1q_u8_x4’ was not declared in this scope; did you mean ‘vld1q_u8’?


benthetechguy:
**Describe the bug**
I am attempting to package xmrig for Debian, and everything is going well so far, except compilation on the `armhf` architecture (armv7l). I ran into this exact issue (#2175) last time I used the normal compilation instructions on Raspbian, Ubuntu, and Arch Linux ARM, and it seemed to have been fixed back then, but is now reappearing.

**To Reproduce**
1. Install Debian sid/unstable on an armv7l machine or VM
2. `sudo apt install build-essential devscripts` (Install package building essentials)
3. `dget -x https://mentors.debian.net/debian/pool/main/x/xmrig/xmrig_6.16.2-1.dsc` (Download the prototype of my package)
4. `cd xmrig-6.16.2 && sudo apt build-dep .` (Install build dependencies)
5. `debuild` (Build the package)
6. See compilation fail

**Expected behavior**
I expected xmrig to compile successfully and the package to build like on the other architectures.

**Required data**
Excerpt of the compilation output ([full output here](https://github.com/xmrig/xmrig/files/7893946/pkgbuild.log)):
```
In file included from /home/ben/xmrig-6.16.2/src/crypto/ghostrider/ghostrider.cpp:57:
/home/ben/xmrig-6.16.2/src/crypto/ghostrider/../../crypto/cn/sse2neon.h: In function ‘uint8x16x4_t _sse2neon_vld1q_u8_x4(const uint8_t*)’:
/home/ben/xmrig-6.16.2/src/crypto/ghostrider/../../crypto/cn/sse2neon.h:363:12: error: ‘vld1q_u8_x4’ was not declared in this scope; did you mean ‘vld1q_u8’?
  363 |     return vld1q_u8_x4(p);
      |            ^~~~~~~~~~~
      |            vld1q_u8
make[3]: *** [src/crypto/ghostrider/CMakeFiles/ghostrider.dir/build.make:303: src/crypto/ghostrider/CMakeFiles/ghostrider.dir/ghostrider.cpp.o] Error 1
```

Spudz76:
What is the gcc version being used for the build?

benthetechguy:
11.2

Spudz76:
Please try with small change of `10` to `11` in `src/crypto/cn/sse2neon.h` line 347:

[...]

PR #2882, opened 2022-01-21
armv7 compilation fix

benthetechguy:
Compilation fails for armv7 on gcc 11, and updating the version in that line fixes it.


SChernykh:
This is not needed for gcc11 on armv8, it will only make it slower there. You need better macro check that turns it on only on armv7.

SChernykh (approved):

xmrig:
@benthetechguy What about https://github.com/xmrig/xmrig/issues/2876#issuecomment-1016931810 is it still present?
Thank you.

benthetechguy:
Yes, but it's unrelated to this PR. They're two separate issues and I'm not
sure how to fix that one.

PR #2927, opened 2022-02-07
Check for required L3CACHE object type in hwloc

Spudz76:
When compiling with hwloc less than v2.x, the compile crashes within GhostRider code due to missing `HWLOC_OBJ_L3CACHE` object type.

Check for the symbol and automatically force `WITH_GHOSTRIDER=OFF` when the detected hwloc does not have it.


mckaygerhard:
after merge this i have this on special OS's : `error: ‘is_trivially_copyable’ is not a member of ‘std’`
```
In file included from /opt/xmrig/src/base/net/stratum/Job.cpp:33:0:
/opt/xmrig/src/base/tools/Alignment.h: In function ‘T xmrig::readUnaligned(const T*)’:
/opt/xmrig/src/base/tools/Alignment.h:33:19: error: ‘is_trivially_copyable’ is not a member of ‘std’
     static_assert(std::is_trivially_copyable<T>::value, "T must be trivially copyable");
                   ^
```

xmrig:
Replaced by https://github.com/xmrig/xmrig/commit/554b60966bdbde1d426a7689eb871c711d9e7666
Thank you.

PR #2935, opened 2022-02-20
Snap created

kz6fittycent:
Initial PR xmrig project

* Snap successfully created and published to snapcraft.io

* Will allow auto-updating of xmrig for users, instead requiring repeated downloads of binaries, or recompiling from source
* When commits made to main repo, builds are automated on snapcraft.io
* Updated README to reflect snap installation instructions

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -3,9 +3,15 @@ adopt-info: xmrig
 summary: XMRig is a high performance, open source, cross platform CPU/GPU miner
 description: |
   XMRig is a high performance, open source, cross platform RandomX, KawPow, 
-  CryptoNight and AstroBWT unified CPU/GPU miner and RandomX benchmark. Official 
-  binaries are available for Windows, Linux, macOS and FreeBSD.
+  CryptoNight and AstroBWT unified CPU/GPU miner and RandomX benchmark. 
   
+  To run, open a terminal:
+    - Ensure you've identified the options you'd like to run
+    - Identify the pool (and port) desired
+    - Use your wallet address
+      - Example: `xmrig -o example.pool.com:4444 -u [wallet address]`
+    - Hit enter and profit
+    
 license: GPL-3.0
 
 base: core20
@@ -15,7 +21,7 @@ compression: lzo
 
 architectures:
   - build-on: amd64
-    run-on: [amd64, armhf, arm64, ppc64el]
+    run-on: [amd64, armhf, arm64, ppc64el, s390x]
 
 assumes:
   - command-chain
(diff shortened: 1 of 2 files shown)


Spudz76:
Should use advanced build using bundled deps, distro dev packages can be unreliable.

kz6fittycent:
I can definitely give that a shot. Building snaps can be pretty tricky.

kz6fittycent:
Okay, that went well enough. Now performing "advanced build". Build is successfully published and running on test machines.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,82 @@
...
+      snapcraftctl set-version "$(git describe --tags | sed 's/^v//' | cut -d "-" -f1)"
+      
+    build-packages:
+      - build-essential 
+      - git
+      - cmake
+      - automake
+      - libtool
+      - autoconf
+      - libuv1-dev
```
Spudz76, line 66: Remove all `-dev` packages, not needed (or wanted) when using bundled.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,82 @@
...
+license: GPL-3.0
+
+base: core20
+grade: stable
+confinement: strict
+compression: lzo
+
+architectures:
+  - build-on: amd64
+    run-on: [amd64, armhf, arm64, ppc64el, s390x]
```
Spudz76, line 24: Technically could have `build-on: [amd64, arm64]`

Seems like run-on should have the same list as build-on, ppc64el and s390x definitely unsupported, etc.

snap/snapcraft.yaml
kz6fittycent: These are build packages, not actually bundled with the snap. Are you saying that the following are not needed during runtime?
```
      - libuv1-dev 
      - libssl-dev 
      - libhwloc-dev
```

snap/snapcraft.yaml
kz6fittycent: This is snapcraft formatting. I am checking into it's capability to run on other arches.

kz6fittycent: Okay, reading this [documentation](https://snapcraft.io/docs/architectures), I will make sure the two architectures you noted above are the only ones; that's assuming that I'm interpreting the language appropriately.

snap/snapcraft.yaml
Spudz76: Correct, bundled libs are equivalent to those three packages.

kz6fittycent:
@Dusit2546 sorry for the delay. Are you asking me to PR to the master branch?

PR #2941, opened 2022-02-21
Update Deps

hilga007:
The current OpenSSL script downloads and builds OpenSSL 1.1.1(x) -- this script downloads and builds the 3.x branch (currently at 3.0.1)


hilga007:
hwloc 2.5.0 --> 2.6.0 
openssl 1.1.1l --> 1.1.1m
libressl 3.0.2 --> 3.4.2
add openssl3.sh to download and build openssl 3.0.2

SChernykh (changes requested):
Why does it have all these OpenCL changes? They're not needed at all.

hilga007:
> Why does it have all these OpenCL changes? They're not needed at all.

Oh CRUD that's a noob alert. I was playing around in master apparently!!!!!!! Was tinkering with spudz' tips for OpenCL on older Intel iGPUs, then just experimenting. Did NOT mean for those to be included! Really just wanted to suggest dep changes! Its almost 2AM so I'll de-junkify later. WHOOPS!!!!!

Issue #3023, opened 2022-04-16
reliabel information about http api ?


turutzi:
Hey guys,

i just tinker a little bit around whit xmirg.
I want to controll my miner over the http api but there isn't anywhere any good or reliabel informations about the api.

Yeah here and there, are some bit of informations but no real list of all commands and controll or request structures.

Would be nice if the http api would be more clear to use.

PR #3030, opened 2022-04-17
Update API Documentation

Spudz76:
Spurred by #3023 and others before

Code changes for clarity and consistency, routing logic consolidated and improved.

Doc changes to fit current reality, add `/json_rpc` endpoint and available methods.


Spudz76:
Also added `json_rpc` method `start` so that `stop` isn't such a dead-end.

Functionally, just calls `onConfigChanged` with the same config as is currently running, which triggers a soft restart of all backends.

PR #3073, opened 2022-06-17
add armv8l architecture

wong-fi-hung:
I'am currently running ubuntu on android terminal emulator for compiling xmrig.
but failed because the different architecture name, but actually is same with aarch64, arm64, or armv8-a.
the app package name is: jackpal.androidterm and can downloaded from playstore.
without adding armv8l architecture to cpu.cmake, I can't compile xmrig on my ubuntu because the gcc keep with error message.
after I'am adding armv8l arch name to cmake/cpu.cmake the compilation procces completed with succesfull.
can you to change your files too...???
because is very helpfully for people like me.
Thankyou...   :D

### DIFF
--- a/cmake/cpu.cmake
+++ b/cmake/cpu.cmake
@@ -29,7 +29,7 @@ else()
 endif()
 
 if (NOT ARM_TARGET)
-    if (CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64|arm64|armv8-a|armv8ls)$")
+    if (CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64|arm64|armv8-a|armv8l)$")
         set(ARM_TARGET 8)
     elseif (CMAKE_SYSTEM_PROCESSOR MATCHES "^(armv7|armv7f|armv7s|armv7k|armv7-a|armv7l)$")
         set(ARM_TARGET 7)
--- a/cmake/cpu.cmake
+++ b/cmake/cpu.cmake
@@ -29,7 +29,7 @@ else()
 endif()
 
 if (NOT ARM_TARGET)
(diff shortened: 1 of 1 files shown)


SChernykh:
armv8l is 32-bit, you'll have very low hashrate. It's better to install Termux from F-droid and compile for 64-bit armv8-a there.

wong-fi-hung:
my device is 64 bit or aarch64 and have 32 bit shared library too.
if run lscpu on termux, the result is aarch64.
but, I'am using terminal other than termux.
so the result is armv8l (ARMV8L).
For spesificly, i'am using xiaomi redmi 4X.
but my ubuntu rootfs is run on 64 bit mode inside the terminal emulator.
and the terminal keep says is the armv8l or 32 bit.
when mining using rx/0 algo.
the randomx hashrate is a normal and can up to 62 H/s if using full 8 threads on my android

this the result:

xmrig: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, BuildID[sha1]=75f2bc50fd3fbd49ce39c1b76f230bf811177066, for GNU/Linux 3.7.0, stripped

benthetechguy:
Your chosen terminal emulator is changing the architecure incorrectly; this is an issue with them and not xmrig. As @SChernykh said, true armv8l is 32 bit and cannot be placed as an alias to "arm64" here.

wong-fi-hung:
> Your chosen terminal emulator is changing the architecure incorrectly; this is an issue with them and not xmrig. As @SChernykh said, true armv8l is 32 bit and cannot be placed as an alias to "arm64" here.

okay thanks all...  :+1: 
i close my pull now and create other idea to my forked repos.
and i will create pull again in other time if possible.

PR #3103, opened 2022-08-05
Support macsmc-ac/online sysfs check for pause-on-battery

### LINKED ISSUE #3102: Xmrig does not detect if the computer is charging (specific hardware and linux kernel)
- hardware: M1 MacBook Pro (2020)
- software: gentoo linux on ARM64/AARCH64, runs [this kernel](https://github.com/AsahiLinux/linux/tree/asahi-5.19-rc7-1) (for hardware support reasons), [manually configured](https://wiki.gentoo.org/wiki/Kernel/Configuration)
- xmrig: emerged from the [`gentoo` repository](https://packages.gentoo.org/packages/net-misc/xmrig)
part of `emerge --info`:
```
=================================================================
                        Package Settings
=================================================================

net-misc/xmrig-6.18.0::gentoo was built with the following:
USE="hwloc ssl -donate (-opencl)"
```

to summarize the problem, here's just a single screenshot with most of the details:

- pause on battery is enabled
- xmrig does not detect if the charger is plugged in or not, or does not respond to the charger state. (`/sys/class/power_supply/macsmc-ac/online`)
- plasma battery status in the taskbar is able to detect if the computer is charging

Spudz76:
Attempt to support/fix #3102


SChernykh:
If I read it right, it's not a standard location for "on battery" status.

Spudz76:
Yes but at the time the `macsmc-battery` sysfs node was not known to me from the information given, so checking AC online is the inverse of on-battery.  If it's AC-offline but still running then obviously it's on battery or it wouldn't be running.

Issue #3136, opened 2022-10-11
How to build a full static executable on MacOS ?


FSOL06:
My goal is to compile XMRIG for ild intel Macs built between 2007 and 2011. I'm running on MacOS 10.5 last release. 

I succeeded - partially - to do it by following the advanced tutorial on the page https://xmrig.com/docs/miner/build/macos

Thanks for that ! 

However, I have a problem. The executable only works on the Mac where the executable was compiled. I get error messages for missing dependencies :

![image](https://user-images.githubusercontent.com/97795464/195136621-85cfdd77-a4e3-4cc4-979b-50fb454d737a.png)

I don't want to install brew & hwloc lib by hand. So I try to build an executable with all the used libraries in static. 

I found an option for CMAKE in the documentation :

![image](https://user-images.githubusercontent.com/97795464/195137068-9277e057-53eb-4d3d-a742-bdd6a5e644a6.png)

Unfortunately, if I use it, it generates a bug at compile time :

![image](https://user-images.githubusercontent.com/97795464/195137765-0b489e25-9706-4804-a792-c902bf7bffbf.png)

How to build a full static executable on MacOS ?

 Thanks for your help.

Spudz76:
Try `scripts/build_deps.sh` to build the exact appropriate versions of the deps, rather than using whatever brew provides (may be incompatible deps versions).

FSOL06:
Where / when I should execute this command ? 

```
1. brew install cmake wget automake libtool autoconf
2. git clone https://github.com/xmrig/xmrig.git
3. mkdir xmrig/build && cd xmrig/scripts
4. ./build.hwloc.sh && cd ../build
5. cmake .. -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl -DHWLOC_INCLUDE_DIR=../scripts/deps/include -DHWLOC_LIBRARY=../scripts/deps/lib/libhwloc.a
6. make -j$(sysctl -n hw.logicalcpu)

```
Between which command or instead of which command ?

FSOL06:
I change line 4 by this : 

`4. ./build_deps.sh && cd ../build
`
This is the result : 

![image](https://user-images.githubusercontent.com/97795464/195172734-28c94e39-5685-4bff-b9b6-4ae0e74d3413.png)

Spudz76:
Replace also line 5:

`cmake .. -DXMRIG_DEPS=../scripts/deps`

You can validate the needed dynamic libs with otool once the binary is built:
[...]

PR #3138, opened 2022-10-17
Make AMD assembly completely optional through WITH_ASM_AMD (default ON)

Spudz76:
Older Apple assembler doesn't understand some AMD opcodes, this allows completely disabling AMD assembly (ryzen/bulldozer) since Apple hardware would never have an AMD anyway.  Only works with UNIX-like platforms, ignored on MSVC (not a problem there anyway).

Also makes the AMD MSRs optional as well (separate configurable `WITH_MSR_ZEN`) for a fully Intel-only executable.

Related to #3136


SChernykh:
> intel Macs built between 2007 and 2011

All these changes to support 0.01% of the user base, not a fair trade I think.

SChernykh:
If it was bigger %, what I would do is find asm functions that don't compile, move them to a separate .s file and just remove them from the build in CMakeLists.txt for these systems, not touching any .h/.cpp files.

SChernykh:
Also, the error in https://github.com/xmrig/xmrig/issues/3136#issuecomment-1279556186 has nothing to do with AMD or AMD Zen assembler code specifically, it's just a `mov	rax, 4389456576511` instruction that works on any CPU, which makes this whole PR pointless.

Spudz76:
IDK, just helping the user.  Who knows why Apple assembler doesn't like that instruction which only appears in the ryzen CN code.

PR #3168, opened 2022-11-21
Add Randomx/xdag (xDagger Coin)

afiniel:
Added rx/xdag ( xDagger Coin )

### DIFF
--- a/src/backend/opencl/cl/cn/algorithm.cl
+++ b/src/backend/opencl/cl/cn/algorithm.cl
@@ -24,6 +24,7 @@
 #define ALGO_RX_SFX         0x72151273
 #define ALGO_RX_KEVA        0x7214116b
 #define ALGO_RX_GRAFT       0x72151267
+#define ALGO_RX_XDAG        0x72151208
 #define ALGO_AR2_CHUKWA     0x61130000
 #define ALGO_AR2_CHUKWA_V2  0x61140000
 #define ALGO_AR2_WRKZ       0x61120000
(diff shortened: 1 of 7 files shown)


src/backend/opencl/cl/cn/algorithm.cl
```diff
@@ -24,6 +24,7 @@
 #define ALGO_RX_SFX         0x72151273
 #define ALGO_RX_KEVA        0x7214116b
 #define ALGO_RX_GRAFT       0x72151267
+#define ALGO_RX_XDAG        0x72151208
```
Spudz76, line 27: Undocumented convention, but the last byte is the ASCII lowercase first letter of the variant name (`s`/`k`/`g`...)

So this should be `0x72151278` because `'x' === 0x78`

src/base/crypto/Algorithm.h
```diff
@@ -78,6 +78,7 @@ class Algorithm
         RX_GRAFT        = 0x72151267,   // "rx/graft"         RandomGRAFT (Graft).
         RX_SFX          = 0x72151273,   // "rx/sfx"           RandomSFX (Safex Cash).
         RX_KEVA         = 0x7214116b,   // "rx/keva"          RandomKEVA (Keva).
+        RX_XDAG         = 0x72151208,   // "rx/xdag"          RandomXDAG (xDagger Coin).
```
Spudz76, line 81: See comment in `algorithm.cl` , this should also be `0x72151278`

SChernykh:
There's no point in this change. It's exactly the same as `rx/0`, so just use `rx/0` in the config.json

Spudz76:
And, that.

afiniel:
Gotcha! There is the changes.

PR #3232, opened 2023-03-25
DaemonClient: new X-Hash-Difficulty HTTP header optimization

moneromooo-monero:
If the caller knows the difficulty of a PoW hash a given nonce yields, it can tell the callee via the X-Hash-Difficulty, which may allow the callee to skip some processing if the difficulty does not meet some criterion.

In my case, a merge mining proxy can know it's pointless trying to submit the nonce to a chain with higher difficulty when the nonce only meets the difficulty for a lower difficulty chain.

### DIFF
--- a/src/base/net/stratum/DaemonClient.cpp
+++ b/src/base/net/stratum/DaemonClient.cpp
@@ -178,7 +178,9 @@ int64_t xmrig::DaemonClient::submit(const JobResult &result)
     m_results[m_sequence] = SubmitResult(m_sequence, result.diff, result.actualDiff(), 0, result.backend);
 #   endif
 
-    return rpcSend(doc);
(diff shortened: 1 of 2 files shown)


moneromooo-monero:
For reference, the use I make of it:
from https://git.townforge.net/townforge/townforge/src/branch/cc/src/merge_mining_proxy/merge_mining_proxy.cc#L856
to https://git.townforge.net/townforge/townforge/src/branch/cc/src/merge_mining_proxy/merge_mining_proxy.cc#L862
which can then avoid a submitblock RPC here:
to https://git.townforge.net/townforge/townforge/src/branch/cc/src/merge_mining_proxy/merge_mining_proxy.cc#L916

SChernykh (approved):

PR #3235, opened 2023-03-27
fix(cuda): receive CUDA loader error on linux too.

0x241F31:
Right now if provider cuda loader fails to load on linux it wouldn't show error, instad log  `CUDA: disabled ((null))`


SChernykh (approved):

SChernykh:
You should've opened this pull request against the dev branch, not master branch.

PR #3239, opened 2023-03-31
Improve .cmd files when run by shortcuts on another drive

koitsu:
### Problem

When using a shortcut on C: drive that points to an xmrig script on another drive (ex. D:), and the shortcut is run as Administrator, the resulting script reports xmrig.exe not found.

### Example scenario

- Shortcut file: `C:\Users\name\Desktop\xmrig.lnk`
- "Target" field in the shortcut points to `D:\xmrig\benchmark_1M.cmd`
- Right-click on desktop shortcut, choose Run as administrator
- Result: `'xmrig.exe' is not recognized as an internal or external command, operable program or batch file.`

### Fix/workaround

As of Windows 10 (possibly 8), you cannot work around this problem using the "Start in" field of shortcuts; it isn't honoured when things are run as Administrator.  This is intentional on Microsoft's part, citing some security-related concern.  Debating this is outside the scope of this PR.  You can Google things like `windows "start in" C:\windows\system32 administrator` and read about it yourself, but https://github.com/microsoft/terminal/issues/7062 talks about it in detail.

Therefore, the proper workaround is use `cd /d` which changes drive letters when given a full path that includes drive letters (which `%~dp0` always does).

This flag has existed since at least Windows 7 (unsure about Vista, XP, or 2K).

Additionally, use double-quotes around `%~dp0` in the case there are spaces anywhere in the pathname.  Refer to the `cd /?` documentation for details about this; it's best to do it because we can't guarantee Command Extensions haven't been turned off by a user.

### Demonstrating the problem and fix:

- Shortcut file: `C:\Users\name\Desktop\test.lnk`
- "Target" points to `D:\xmrig\test.cmd`
- `D:\xmrig\test.cmd` contents:

```
@echo off
echo cwd 1: %cd%
cd %~dp0
echo cwd 2: %cd%
cd /d "%~dp0"
echo cwd 3: %cd%
pause
```

- Right-click on desktop shortcut, choose Run as administrator
- Results are below -- cwd 2 might surprise you:

```
cwd 1: C:\Windows\system32
cwd 2: C:\Windows\system32
cwd 3: D:\xmrig
Press any key to continue . . .
```

### DIFF
--- a/scripts/benchmark_10M.cmd
+++ b/scripts/benchmark_10M.cmd
@@ -1,4 +1,4 @@
 @echo off
-cd %~dp0
+cd /d "%~dp0"
 xmrig.exe --bench=10M --submit
 pause
--- a/scripts/benchmark_1M.cmd
+++ b/scripts/benchmark_1M.cmd
@@ -1,4 +1,4 @@
 @echo off
-cd %~dp0
+cd /d "%~dp0"
 xmrig.exe --bench=1M --submit
 pause
--- a/scripts/pool_mine_example.cmd
+++ b/scripts/pool_mine_example.cmd
@@ -15,6 +15,6 @@
 :: Choose pools outside of top 5 to help Monero network be more decentralized!
 :: Smaller pools also often have smaller fees/payout limits.
 
-cd %~dp0
+cd /d "%~dp0"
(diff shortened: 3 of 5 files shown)


SChernykh (approved):

SChernykh:
This PR should've been opened against `dev` branch.

koitsu:
@SChernykh I wasn't aware of said branch; project has no CONTRIBUTING.md file, nothing in README.md, or in the docs/ directory mentioning it either.  I'm used to following https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors

Would you like me to resubmit the change + PR against `dev` branch?

SChernykh:
Yes, it would be nice if you PR against `dev`. Thank you!

PR #3278, opened 2023-05-29
Allow using system libfmt

jeandudey:
Some distributions such as GNU Guix don't allow using bundled dependencies and code, so I'm adding the option to use a system version of `libfmt` instead of using the bundled one.

Had to nest `fmt` and `CL` in another sub-directory of the same name to avoid accidental inclusion of the `fmt` headers.


SChernykh:
> don't allow using bundled dependencies and code

How do they not allow it? It just doesn't compile? What is the error?

jeandudey:
>How do they not allow it? It just doesn't compile? What is the error?

@SChernykh the GNU Guix project generally tries to remove bundled dependencies from packages if it's available already, for example, [fmt](https://packages.guix.gnu.org/packages/fmt) is already packaged and it the library tests are run after building it, libraries are also un-bundled for auditability and to avoid re-building code that's already built.

For example, on the package [definition](https://github.com/guix-mirror/guix/blob/7b400e7f8751e6b0cc6e66d3f7ecfb7f5bd51309/gnu/packages/finance.scm#L2199) the [hwloc](https://github.com/guix-mirror/guix/blob/7b400e7f8751e6b0cc6e66d3f7ecfb7f5bd51309/gnu/packages/finance.scm#L2216) library is unbundled as it is available on GNU Guix and xmrig supports building without it being on the `src/3rdparty` directory, other libraries such as OpenCL headers and libfmt are also available but require changes in xmrig first to avoid patching large chunks of the source code in the package definition.

This is done so that GNU Guix creates a source archive of the xmrig source code but doesn't contain bundled code in the `src/3rdparty` directory.

SChernykh:
In this case it should have CMakeLists option for libfmt, just like hwloc has `WITH_HWLOC`. Also, the PR should be made to `dev` branch, not `master`.

jeandudey:
Added the option in https://github.com/xmrig/xmrig/pull/3278/commits/c81d154f3bf7a60a5421ab17dd9d10a3e6a29b39.

(made fixup commit to squash once the review is done)

PR #3342, opened 2023-10-08
Update cn_main_loop.asm

JacksonZ03:
I was scanning the code and found this line to be missing. Not sure if this was a mistake or if it was intentionally left out?


SChernykh (approved):

SChernykh:
This file is only used when compiling with Visual Studio. I guess it's not necessary to declare a symbol PUBLIC there to make it visible, because compilation didn't break.

PR #3420, opened 2024-02-11
Fix segfault in HTTP API rebind

Daviey:
Previously with HTTP API enabled on brenchmarking run, it is possible to cause a segfault due to an issue handling the m_httpd pointer and rebinding.

  - Initialize m_httpd to nullptr to indicate when it's not in use.
  - Safely delete m_httpd in Api's destructor to prevent use-after-free issues.
  - Add checks to ensure m_httpd is not nullptr before usage in start, stop, and tick methods.
  - Log errors for HTTP server start failures to aid in debugging.

Fixes MoneroOcean/xmrig#120

### DIFF
--- a/src/base/api/Api.cpp
+++ b/src/base/api/Api.cpp
@@ -39,6 +39,7 @@
 
 
 #include <thread>
+#include <iostream>
 
 
 namespace xmrig {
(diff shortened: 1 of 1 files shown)


SChernykh (approved):

xmrig:
Merged into [dev](https://github.com/xmrig/xmrig/commit/2c9c40d623544f7adbd8e9de4875a97da799fc5e) branch.
Thank you.

PR #3427, opened 2024-02-22
solo_mine_example.cmd

Yaxxir:


SChernykh (changes requested):
You shouldn't be changing user examples with your own wallet addresses

Issue #3440, opened 2024-03-11
Getting error on raspberry pi 5


yashkathe:
```
[ 88%] Building CXX object CMakeFiles/xmrig.dir/src/crypto/randomx/soft_aes.cpp.o
/home/yashkathe/xmrig/src/crypto/randomx/randomx.cpp: In member function ‘void RandomX_ConfigurationBase::Apply()’:
/home/yashkathe/xmrig/src/crypto/randomx/randomx.cpp:274:38: error: ‘randomx::JitCompilerA64’ has not been declared
  274 | #define JIT_HANDLE(x, prev) randomx::JitCompilerA64::engine[k] = &randomx::JitCompilerA64::h_##x
      |                                      ^~~~~~~~~~~~~~
/home/yashkathe/xmrig/src/crypto/randomx/randomx.cpp:285:37: note: in expansion of macro ‘JIT_HANDLE’
  285 |         for (; k < freq_sum; ++k) { JIT_HANDLE(x, prev); }
      |                                     ^~~~~~~~~~
/home/yashkathe/xmrig/src/crypto/randomx/randomx.cpp:291:9: note: in expansion of macro ‘INST_HANDLE’
  291 |         INST_HANDLE(IADD_RS, NULL);
      |         ^~~~~~~~~~~
/home/yashkathe/xmrig/src/crypto/randomx/randomx.cpp:274:76: error: ‘randomx::JitCompilerA64’ has not been declared
  274 | #define JIT_HANDLE(x, prev) randomx::JitCompilerA64::engine[k] = &randomx::JitCompilerA64::h_##x
      |                                                                            ^~~~~~~~~~~~~~
/home/yashkathe/xmrig/src/crypto/randomx/randomx.cpp:285:37: note: in expansion of macro ‘JIT_HANDLE’
  285 |         for (; k < freq_sum; ++k) { JIT_HANDLE(x, prev); }
      |                                     ^~~~~~~~~~
/home/yashkathe/xmrig/src/crypto/randomx/randomx.cpp:291:9: note: in expansion of macro ‘INST_HANDLE’
  291 |         INST_HANDLE(IADD_RS, NULL);

```

SChernykh:
You're probably using 32-bit OS. You must compile XMRig in 64-bit.

yashkathe:
I tried to run it on Raspberry PI OS (64 bit) with desktop

But once I tried to install normal Raspberry pi OS without desktop it somehow worked
Thank You

PR #3462, opened 2024-04-14
RandomX: correct memcpy size for JIT initialization

SChernykh:
No buffer overflow, better fix for `_FORTIFY_SOURCE`. Replaces #3450 and #3461.


d4f5409d:
I think this helped to fix some aspects of #3440 so thank you 💞

PR #3463, opened 2024-04-16
chore: fix typos

xiaoxianBoy:
fix some typos


RobQuistNL:
these are copied files from 3rd parties - you should fix it in those repo's and then pull that new version in here

PR #3464, opened 2024-04-18
chore: fix some typos in comments

growfrow:


growfrow:
Anything need I to do? If have, Please feel free to tell me.

RobQuistNL:
these are copied files from 3rd parties - you should fix it in those repo's and then pull that new version in here

PR #3475, opened 2024-05-03
Create a sleep mechanism for slowing down mining process

### LINKED ISSUE #1322: How to change intensity and worksize in v5.0.1?
How do you change these parameters in the latest version to bypass the default settings for the NVIDIA card?

James4Ever0:
This pull request provides a configurable environmental variable called `XMRIG_SLEEP_NANOSECONDS` which value is default to zero.

One can configure it to slow down the mining process, effectively eliminate the issue of utilizing a single thread 100% at all the time, thereby increase the overall system stability and lower the thermo pressure.

Fix #1322

### DIFF
--- a/src/core/Miner.cpp
+++ b/src/core/Miner.cpp
@@ -457,6 +457,8 @@ const std::vector<xmrig::IBackend *> &xmrig::Miner::backends() const
 
 xmrig::Job xmrig::Miner::job() const
 {
+    std::this_thread::sleep_for(std::chrono::nanoseconds(sleepNanoSeconds));
+    
     std::lock_guard<std::mutex> lock(mutex);
 
     return d_ptr->job;
--- a/src/core/Miner.cpp
+++ b/src/core/Miner.cpp
@@ -380,6 +380,14 @@ public:
 xmrig::Miner::Miner(Controller *controller)
     : d_ptr(new MinerPrivate(controller))
 {
+
+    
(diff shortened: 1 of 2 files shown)


SChernykh:
This is a weird place to put `sleep` into. If anything, it should be in the main CPU mining loop, next to already existing `yield` (with the similar purpose): https://github.com/xmrig/xmrig/blob/master/src/backend/cpu/CpuWorker.cpp#L357

James4Ever0:
Now I have made some progress here. I make the dataset preparation process single threaded in ASM mode (because I cannot figure out how to sleep in assembly code in PIE compilation mode), throttled in non-ASM mode but taking ages to prepare so therefore out of major concern, and finally throttled hash functions.

SChernykh:
Honestly, this PR is a bunch of hack on top of each other, and there is 0% probability it will be merged. It's fine for your own purposes, I guess.

SChernykh:
https://xmrig.com/docs/miner/command-line-options
`--randomx-init=N	thread count to initialize RandomX dataset`

James4Ever0:
That being said, I think it is possible to create some better mechanism for controlling thread occupation, by first measuring the average time to perform a single loop in hash function, then sleep accordingly to the mearured time, so that the user will only have to provide a value ranging from 1 to 99 in percentage, simplifying the guesswork.

PR #3486, opened 2024-05-30
hwloc: Clarify reason for failed binding ("can't bind memory")

Spudz76:
Memory binding error message can be mysterious, add a bit of info as to which call actually failed to assist in solving the problem.


src/backend/cpu/platform/HwlocCpuInfo.cpp
```diff
@@ -191,16 +192,25 @@ xmrig::HwlocCpuInfo::~HwlocCpuInfo()
...
+int8_t xmrig::HwlocCpuInfo::membind(hwloc_const_bitmap_t nodeset)
 {
     if (!hwloc_topology_get_support(m_topology)->membind->set_thisthread_membind) {
-        return false;
+        return MEMBIND_FAIL_SUPP;
     }
 
 #   if HWLOC_API_VERSION >= 0x20000
-    return hwloc_set_membind(m_topology, nodeset, HWLOC_MEMBIND_BIND, HWLOC_MEMBIND_THREAD | HWLOC_MEMBIND_BYNODESET) >= 0;
+    static int rv = hwloc_set_membind(m_topology, nodeset, HWLOC_MEMBIND_BIND, HWLOC_MEMBIND_THREAD | HWLOC_MEMBIND_BYNODESET);
```
SChernykh, line 202: `static` is not needed here

src/crypto/rx/RxNUMAStorage.cpp
```diff
@@ -42,20 +42,20 @@ constexpr size_t oneMiB = 1024 * 1024;
...
-static bool bindToNUMANode(uint32_t nodeId)
+static int8_t bindToNUMANode(uint32_t nodeId)
 {
     auto node = hwloc_get_numanode_obj_by_os_index(Cpu::info()->topology(), nodeId);
     if (!node) {
-        return false;
+        return MEMBIND_FAIL_NODE;
     }
 
     if (Cpu::info()->membind(node->nodeset)) {
```
SChernykh, line 52: This check must be changed too

Spudz76:
Not sure how I messed that up, been running this patch for year(s) but I also don't have anything to test on where it fails.

src/backend/cpu/platform/HwlocCpuInfo.cpp
```diff
@@ -191,16 +192,25 @@ xmrig::HwlocCpuInfo::~HwlocCpuInfo()
...
+        return MEMBIND_FAIL_SUPP;
     }
 
 #   if HWLOC_API_VERSION >= 0x20000
-    return hwloc_set_membind(m_topology, nodeset, HWLOC_MEMBIND_BIND, HWLOC_MEMBIND_THREAD | HWLOC_MEMBIND_BYNODESET) >= 0;
+    int rv = hwloc_set_membind(m_topology, nodeset, HWLOC_MEMBIND_BIND, HWLOC_MEMBIND_THREAD | HWLOC_MEMBIND_BYNODESET);
+    int error = errno;
+
+    if (rv < 0) {
+        printf("hwloc_set_membind() error: %s\n", strerror(error));
```
SChernykh, line 206: You should use `LOG_WARN` instead, otherwise it won't be logged properly.

SChernykh: That line is debugging code, it's inside `#ifdef XMRIG_HWLOC_DEBUG`.

Spudz76:
Caught it actually working:

image

PR #3489, opened 2024-06-01
updated default DNS TTL to 300

cyboerg42:
This pull-request will change the default DNS TTL of 30s to 300s.

It'll take some slack of DNS, but this can only stay as a hot-fix. It should read the TTL from the NS and use it, that the pool/DNS providers can adapt their settings for their needs.  Some will need larger, some will need smaller TTLs. 300 should be a sane default for now.

### DIFF
--- a/src/base/net/dns/DnsConfig.h
+++ b/src/base/net/dns/DnsConfig.h
@@ -44,7 +44,7 @@ public:
 
 private:
     bool m_ipv6     = false;
-    uint32_t m_ttl  = 30U;
+    uint32_t m_ttl  = 300U;
 };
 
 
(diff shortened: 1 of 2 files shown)


SChernykh (approved):

xmrig:
The name of this option may be confusing, but this option sets TTL for the internal cache and is not supposed to know anything about the DNS TTL.
The cache is used to rate limit DNS requests, with a default value of a maximum of 1 request per 30 seconds per domain. This option is most useful for proxy, which can create a lot of requests. Five minutes is too high for the default value. Even 30 seconds might be high, so this option is configurable.
Thank you.

PR #3491, opened 2024-06-03
Call `uv_loop_fork` to redecorate the default loop after `fork()`

### LINKED ISSUE #3455: Personal log doesn't work with --background
**Describe the bug**
Personal log doesn't work with --background

**To Reproduce**
```
sudo ./xmrig --algo=rx/0 --url=rx.unmineable.com:3333 --user=$coin:$address.$worker --keepalive --threads=2 --background --log-file=/home/user/my_log.log
```
No logs in "my_log.log
If I delete "--background" the records appear in the terminal and in the "my_log.log"

```
user@mst:~$ ls -l my_log.log
-rwxrwxrwx 1 user user 0 abr  3 09:17 my_log.log
```

**Expected behavior**
That the logs be saved in the custom log `my_log.log` and can use the `--background` option, so that nothing appears in the console

**Required data**
 - XMRig version 6.21.2
 - OS: Ubuntu 22.04
 - CPU Intel Xeon

**Workaround**

replace personal log with `--syslog`

```
sudo ./xmrig --algo=rx/0 --url=rx.unmineable.com:3333 --user=$coin:$address.$worker --keepalive --threads=2 --background --syslog
```

Spudz76:
Fixes #3455 

[libuv docs](https://docs.libuv.org/en/v1.x/loop.html#c.uv_loop_fork) basically say avoid using `fork()`, but if it must be done then you have to use this experimental function to fixup some things in the child, unless you make new loops (aka not use the default loop at all post-fork).

### DIFF
--- a/src/App.cpp
+++ b/src/App.cpp
@@ -61,12 +61,12 @@ int xmrig::App::exec()
         return 2;
     }
 
-    m_signals = std::make_shared<Signals>(this);
-
     int rc = 0;
     if (background(rc)) {
         return rc;
     }
+
+    m_signals = std::make_shared<Signals>(this);
 
     rc = m_controller->init();
     if (rc != 0) {
--- a/src/App_unix.cpp
+++ b/src/App_unix.cpp
@@ -27,6 +27,7 @@
 #include <csignal>
 #include <cerrno>
 #include <unistd.h>
+#include <uv.h>
 
 
 #include "App.h"
@@ -52,6 +53,8 @@ bool xmrig::App::background(int &rc)
 
         return true;
     }
+
+    uv_loop_fork(uv_default_loop());
 
     i = setsid();
 


SChernykh:
I checked the code flow, and the only place where `uv_default_loop` is used before forking, is this one: https://github.com/xmrig/xmrig/blob/dev/src/App.cpp#L64

Can you move it a few lines down (after `if (background(rc))` check) and see if it works? This should make the whole thing more safe.

Spudz76:
According to the docs, using `uv_default_loop()` anywhere after forking will also be broken in "undefined" ways, without calling this fixer function (which is experimental and considered a hack and subject to breakage in future versions).  Basically amazing if the logging is the only thing that doesn't quite work right, but it does appear to connect and run the backends so most things must work (by luck).

I think you mean move it out into the global `App.cpp` however this only does anything on Unix anyway, it is unsupported on Windows.

SChernykh:
No, I mean your fix + this one line moved should be enough.

SChernykh:
This is the change I'm talking about: https://paste.debian.net/hidden/734a8381/

Spudz76:
Yeah I figured that out after a re-read of what you said.  Done.  Testing.

Spudz76:
Appears to work, the log is being written to and flushed properly, connected and is receiving jobs and shows nonzero hashrate statuses, and I waited for at least one accept.  Can't test OpenCL or other places `uv_default_loop()` is used but assume they aren't broken worse than they may already have been.  This was with CPU backend only.

Spudz76:
Kill even with `-INT` did not exit with the stopping messages, it just cuts off.  Not sure if that's normal or not, I thought it logs the exiting when non-background.

Spudz76:
Confirmed when foreground and killed from another shell (not ctrl-c) it logs the SIGINT and exit.

Spudz76:
Perhaps fixing it completely properly with [something like this](https://nikhilm.github.io/uvbook/processes#detaching-processes) instead of `fork()` would help, I'm not sure.  This is literally my first day bothering with how to use libuv, lol.  I also note how the default loop handle is stuffed into a variable instead of called in the child, not sure if that's magical somehow.  But if the handle has to be carried around it might as well be a non-default loop and then replace all calls to uv_default_loop with that pointer.

Spudz76:
Possibly related: https://github.com/libuv/libuv/issues/3483

The patch was merged at some point, checking if its in the bundled version now.

But it appears to make children ignore signals even if the parent exited, rather than making the "main child" react to them when the parent goes away.

SChernykh (approved):

Spudz76:
PR #3492 also fixes the original problem and avoids using the `uv_loop_fork()` function which is likely a better idea.

PR #3492, opened 2024-06-04
Throw away any existing uv_default_loop, recreate after fork

### LINKED ISSUE #3455: Personal log doesn't work with --background

Spudz76:
(also) Fixes #3455 

This is an alternate approach to the problem, it does not function any differently however might be cleaner.  Simply destroy any existing default loop to cause a new one to be created on the next `uv_default_loop()` call (in Signals constructor).  This completely avoids using the experimental/non-guaranteed `uv_loop_fork()` function.

### DIFF
--- a/src/App.cpp
+++ b/src/App.cpp
@@ -61,12 +61,13 @@ int xmrig::App::exec()
         return 2;
     }
 
-    m_signals = std::make_shared<Signals>(this);
-
     int rc = 0;
     if (background(rc)) {
         return rc;
     }
+
+    uv_loop_close(uv_default_loop());
+    m_signals = std::make_shared<Signals>(this);
 
     rc = m_controller->init();
     if (rc != 0) {
--- a/src/App.h
+++ b/src/App.h
@@ -33,6 +33,7 @@
 
 
 #include <memory>
+#include <uv.h>
 
 
 namespace xmrig {


xmrig:
Is `uv_loop_close` really needed? It does not look right to close the freshly created default loop and create it again after. Also, the addition of `#include <uv.h>` is not required in `App.h`.
Thank you.

SChernykh:
`uv_loop_close` will be a no-op at best. If any libuv handle was created at this point, `uv_loop_close` will fail and return `UV_EBUSY` error. If no handles were created, there is no point in closing the loop.

Spudz76:
I agree but it works.  So it must be doing something.

PR #3585, opened 2024-11-15
Fix Typo --> adl_structures.h

YuZekai:


SChernykh (approved):

SChernykh (commented):
This is a 3rd-party code, so it will not be merged to this repository directly. It should be fixed by updating the entire AMD ADL SDK - assuming that it's fixed in the new version.

PR #3661, opened 2025-05-29
Fix type of `keccak_f800`

ybh1998:
The type of `keccak_f800` should be `void` instead of `uint64_t`, which causes kernel compilation error on some platform. OpenCL headers are also regenerated.


SChernykh (approved):

ybh1998:
I have created two PRs. One is to the `master` branch, and the other is to the `dev` branch. Please verify which one is continent for you and I will close the other.

xmrig:
`dev branch is correct.
Thank you.

PR #3665, opened 2025-06-11
Tweaked autoconfig for AMD CPUs with < 2 MB L3 cache per thread

SChernykh:
For some reason, AMD CPUs can use only half of the exclusive L2/L3 cache combo efficiently.
Tested on Ryzen 5 8600G:

|Version|Hashrate | Threads used
|-|-|-|
|XMRig v6.22.2 | 6100 h/s | 8
|XMRig v6.22.3 | 5980 h/s | 11
|XMRig v6.22.3 with this fix | 6250 h/s | 9

### DIFF
--- a/src/backend/cpu/platform/HwlocCpuInfo.cpp
+++ b/src/backend/cpu/platform/HwlocCpuInfo.cpp
@@ -322,7 +322,8 @@ void xmrig::HwlocCpuInfo::processTopLevelCache(hwloc_obj_t cache, const Algorith
 
             if (L3_exclusive) {
                 if (vendor() == VENDOR_AMD) {
-                    extra += std::min<size_t>(l2->attr->cache.size, scratchpad);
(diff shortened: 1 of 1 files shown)


NulliferBones:
This update has made hash way worse on rx/0 on my 1700x

SChernykh:
With this update, it should use 8 threads on 1700x and have the same hashrate as older versions of XMRig. How many threads does it use on your 1700X, and what are you comparing to (hashrate, number of threads)?

NulliferBones:
Previously, my 1700x was using 8 threads and getting about 5200 hash. Now with this update, obviously re benched, and its now using 10 threads and getting 4200 hash. Other algos have seen a performance boost. But rx/0 has lost about 1k.

NulliferBones:
To be clear im talking about 6.23.0

MoneroArbo:
I can confirm 6.23.0 reduces the hash rate on my 6800H from ~5.4k (8 threads) to ~5k (9 threads)

PR #3707, opened 2025-09-13
Fix undefined behavior in keccak.cpp

MoneroOcean:
Fix strict-aliasing issue by introducing uint64_t_alias type

Changed all casts from (uint64_t *) to (uint64_t_alias *) to prevent undefined behavior due to strict-aliasing rules.

According to the standard, an object shall only be accessed through an lvalue of a compatible type, or a few exceptions (char types,  or types with similar representation). Accessing a uint64_t object through a different pointer type like void* or another unrelated type is undefined behavior. Introducing uint64_t_alias ensures type-punning without violating these rules.

This ensures safer type-punning while keeping the original logic intact.

### DIFF
--- a/src/base/crypto/keccak.cpp
+++ b/src/base/crypto/keccak.cpp
@@ -166,6 +166,7 @@ void xmrig::keccakf(uint64_t st[25], int rounds)
 // compute a keccak hash (md) of given byte length from "in"
 typedef uint64_t state_t[25];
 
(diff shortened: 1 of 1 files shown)


SChernykh:
This will not compile on MSVC compiler.

MoneroOcean:
Put fix that should help (not tested it under MSVC yet).

PR #3715, opened 2025-09-28
Fix cryptonight crash when using softaes on arm

1640675651:
Cryptonight related algorithms fails to start on arm devices with softaes (e.g. rpi 3, 4). It prints softaes failed. [](https://github.com/MoneroOcean/xmrig/issues/128.) Although the issue is from the MoneroOcean fork, the official version also has this problem. This seems due some unsafe pointer casting in the soft_aesenc function, which causes wrong result on arm platforms using SSE2Neon. Use union to reinterpret __m128i solves the issue.

### DIFF
--- a/src/crypto/cn/soft_aes.h
+++ b/src/crypto/cn/soft_aes.h
@@ -89,12 +89,33 @@
 alignas(16) const uint32_t saes_table[4][256] = { saes_data(saes_u0), saes_data(saes_u1), saes_data(saes_u2), saes_data(saes_u3) };
 alignas(16) const uint8_t  saes_sbox[256] = saes_data(saes_h0);
 
+typedef union {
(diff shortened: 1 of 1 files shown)


SChernykh:
This needs to be tested on x86 and different compilers too. Because it basically replaces one undefined behavior with another (reinterpret_cast to an unrelated type and union casts are both UB in C++).

Or, even better, use the "working but arm specific" code with `#ifdef XMRIG_ARM` macro. Then the change will only affect arm builds.

1640675651:
Using #ifdef XMRIG_ARM now.

SChernykh:
Looks good, you just need to remove the commented out code because there's no need to have it now.

1640675651:
Removed comments.

PR #3717, opened 2025-10-03
replace CMAKE_SOURCE_DIR with CMAKE_CURRENT_SOURCE_DIR

bsolomenco:
this will fix CMakeLists.txt/CMAKE_MODULE_PATH: use CMAKE_CURRENT_SOURCE_DIR instead CMAKE_SOURCE_DIR because "cmake" subdir is relative to CMAKE_CURRENT_SOURCE_DIR

The issue can be observed when trying to include xmrig's dir in a "parent" CMake project

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -241,14 +241,14 @@ target_link_libraries(${CMAKE_PROJECT_NAME} ${XMRIG_ASM_LIBRARY} ${OPENSSL_LIBRA
 
 if (WIN32)
     if (NOT ARM_TARGET)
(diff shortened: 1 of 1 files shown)


SChernykh:
If I understand it correctly, it only matters if XMRig's CMakeLists.txt itself is included into some upper level CMakeLists.txt of another project.

bsolomenco:
> If I understand it correctly, it only matters if XMRig's CMakeLists.txt itself is included into some upper level CMakeLists.txt of another project.

That's right. I tried to create a "parent" CMake project that builds xmrig + dependencies on Windows.
I found CMAKE_SOURCE_DIR also in other places in xmrig, I'll add another commit to this PR to replace CMAKE_SOURCE_DIR with CMAKE_CURRENT_SOURCE_DIR

PR #3724, opened 2025-10-21
RISC-V Intergration

Slayingripper:
# Tested on: Ky X1 8-core RISC-V AI CPU (via ORANGE PI RV2)

image

I  have successfully integrated initial RISC-V support and conducted preliminary testing on an ORANGE PI RV2 board. With initial optimisations, the board achieved 36 H/s. While this performance is modest, it establishes a functional foundation and opens significant opportunities for testing and optimisation on higher-end RISC-V boards and platforms in the future.

I've recompiled on X86 and a Raspberry Pi 3 and 5 to make sure everything is still compatible.  

## Build System Updates (CMake)

The following changes were made to the CMake build scripts to enable cross-compilation and native compilation for RISC-V:

1. Architecture Detection: Added RISC-V detection logic within cmake/cpu.cmake, cmake/flags.cmake, and other critical build scripts.

2. Aggressive Optimisations: Implemented new compiler flags specifically for RISC-V, targeting aggressive optimisation (-march=rv64gcv_zba_zbb_zbc_zbs, loop unrolling, vectorisation, etc.).

3. Feature Exclusion: Disabled all x86-specific instruction set features (SSE, AVX, VAES, MSR) to ensure successful compilation on RISC-V targets.

##  Source Code Adaptations

Significant modifications were required across the CPU backend, crypto engine, and memory management layers.

1. CPU Backend

- Architecture-Specific Files: Added architecture-specific source files (lscpu_riscv.cpp) and updated existing ones (BasicCpuInfo_arm.cpp) to handle RISC-V architecture details.
 
- Runtime Feature Detection: Implemented custom logic for runtime detection of crypto and vector extensions by parsing /proc/cpuinfo.

2. Crypto Engine

- Vector Intrinsic Compatibility: Introduced new compatibility layers (sse2rvv.h, sse2rvv_optimized.h, sse2rvv_scalar_backup.h) to map x86 SSE intrinsics to RVV (RISC-V Vector) intrinsics.

- RISC-V Crypto Support: Added conditional support for native RISC-V crypto extensions (AES, SHA2, SHA3, bit manipulation) via riscv_crypto.h, including compile-time and runtime detection.

- Cleanup: All x86- and ARM-specific code is now conditionally excluded when compiling for RISC-V.

3. RandomX Optimisations

- Dataset Initialisation: Added new dataset initialisation logic specifically for RISC-V (RxDataset_riscv.h), featuring adaptive threading and cache-aware memory operations.

- Temporary AES Fallback: Forced the use of software AES on RISC-V until native hardware intrinsics are fully stable and integrated.

- Vectorised memory and arithmetic operations using RVV when the extension is detected.

4. Memory and Versioning

- Memory Management: Added RISC-V optimised routines for memory barriers, atomic operations, prefetching, and aligned memory access (riscv_memory.h).

5.  Performance and Tuning

Aggressive compiler optimisations, including vector extensions, bit manipulation, loop unrolling, and math optimisations, are enabled by default to achieve maximum hashing performance. 

While this helped the board go from 14 H/s to 33 H/s to 36 H/s I am not sure how this will work out on other CPUs


src/backend/cpu/platform/BasicCpuInfo_arm_unix.cpp
```diff
@@ -42,13 +42,32 @@ namespace xmrig {
...
+#endif
+
 
 } // namespace xmrig
 
 
 void xmrig::BasicCpuInfo::init_arm()
 {
-#   if __ARM_FEATURE_CRYPTO
+#   ifdef XMRIG_RISCV
```
SChernykh: Same. Please move the code to BasicCpuInfo_riscv.cpp

src/crypto/cn/sse2rvv.h
```diff
@@ -0,0 +1,743 @@
...
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * SSE to RISC-V Vector (RVV) optimized compatibility header
+ * Provides both scalar fallback and vectorized implementations using RVV intrinsics
+ */
```
SChernykh: Please credit the original authors here

src/crypto/riscv/riscv_rvv.h
```diff
@@ -0,0 +1,237 @@
+#ifndef XMRIG_RISCV_RVV_H
```
SChernykh: No credits here

SChernykh (changes requested):
Good initial work! I guess you just went the path of "fix compile errors" until it compiled and passed tests?

Please, for each of the new files, add the correct credits at the top (either for the original authors, or for yourself if you wrote the code). Ideally, also add the links to where you took the code from - it will make maintaining it much easier.

Slayingripper:
Will do.

Pretty much, but I also needed to make some additions so that the correct libraries are loaded and that it actually works "properly".

Once I confirmed that RANDOMX compiles and runs, I was sure I could figure out some way to get XMRig to compile for RISC-V. 

I mean, Bitmain made it work somehow with the X5. 

> Good initial work! I guess you just went the path of "fix compile errors" until it compiled and passed tests?
> 
> Please, for each of the new files, add the correct credits at the top (either for the original authors, or for yourself if you wrote the code). Ideally, also add the links to where you took the code from - it will make maintaining it much easier.

SChernykh:
As soon as this PR is merged, I will add the RISC-V JIT compiler from the reference RandomX repository - it shouldn't be that hard now. I can only test it with qemu though, so I would appreciate if you test it once it's added.

cmake/flags.cmake
```diff
@@ -25,9 +25,32 @@ if (CMAKE_CXX_COMPILER_ID MATCHES GNU)
...
+        add_definitions(-DHAVE_ROTR)
     elseif (ARM_TARGET EQUAL 7)
         set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=armv7-a -mfpu=neon -flax-vector-conversions")
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=armv7-a -mfpu=neon -flax-vector-conversions")
+        add_definitions(-DHAVE_ROTR)
+    elseif (XMRIG_RISCV)
+        # RISC-V-specific optimizations for maximum performance
+        # Use vector extensions (V) and bit manipulation extensions (Zba, Zbb, Zbc, Zbs)
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=rv64gcv_zba_zbb_zbc_zbs")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=rv64gcv_zba_zbb_zbc_zbs")
```
SChernykh: This crashes with `Illegal instruction` in my qemu. Use this as the reference: https://github.com/tevador/RandomX/blob/master/CMakeLists.txt#L185-L207

It should compile and run with `rv64gc` only, even if the performance is lower. Of course you can enable additional extensions when compiling on the real CPU (using the `try_run` function from there).

SChernykh:
One more thing I noticed: it reports CPU as 32-bit because you forgot to fix `is64bit()` function in `src/backend/cpu/interfaces/ICpuInfo.h`

cmake/cpu.cmake
```diff
@@ -29,6 +29,13 @@ else()
     set(WITH_VAES OFF)
 endif()
 
+# Disable x86-specific features for RISC-V
+if (XMRIG_RISCV)
```
SChernykh: This uses `XMRIG_RISCV` before it's actually set.

SChernykh:
image
It works in qemu with `rv64gc` only, that's good news

Slayingripper:
# Addressed the comments 
1.  ``argon2 SSE2 flag`` - Changed from /arch:SSSE3 to empty string
2.   ``BasicCpuInfo.cpp guards (line 26)`` - Removed #if !defined(XMRIG_RISCV)
3.  `` BasicCpuInfo.cpp guards (line 75) `` - Removed duplicate guard
4. `` BasicCpuInfo_riscv.cpp`` - Created separate RISC-V implementation file
5. Move code from ARM files - All RISC-V code now in`` BasicCpuInfo_riscv.cpp``
6. `` lscpu_riscv.cpp ``credits - Added proper copyright with author names
7.  ``cpu.cmake`` - Properly includes BasicCpuInfo_riscv.cpp for RISC-V builds
8. ``sse2rvv.h`` credits - Added copyright and reference to sse2neon.h
9. ``riscv_rvv.h`` credits - Added proper copyright header
10.  cmake flags - Changed to rv64gc base (qemu-compatible)
11.  ``is64bit()`` function - Added RISC-V 64-bit detection
12. XMRIG_RISCV order - NEW FIX: Moved RISC-V detection earlier in cpu.cmake
14. ``cpu.cmake ``: Moved RISC-V architecture detection to line 24 (before it's first used on line 37)

I've recompiled it on the RV2 and run it with no issues. 
There is probably a lot of room for improvement, but it's a start.

SChernykh:
Almost everything is good, you just need to clean up now:

- `.github/workflows/CompileX86.yml` - please remove it because XMRig uses its own build servers
- `BasicCpuInfo_arm.cpp, BasicCpuInfo_arm_unix.cpp` - revert changes in these two files because they were moved to BasicCpuInfo_riscv.cpp
- `.gitignore` - no changes needed there (you can keep those changes locally for your convenience, just not in this PR)

SChernykh:
I have almost finished porting the RISC-V JIT compiler from https://github.com/tevador/RandomX/ so once this PR is merged, I'll make a PR with the JIT compiler and you will be able to test it.

Slayingripper:
> Almost everything is good, you just need to clean up now:
> 
>     * `.github/workflows/CompileX86.yml` - please remove it because XMRig uses its own build servers
> 
>     * `BasicCpuInfo_arm.cpp, BasicCpuInfo_arm_unix.cpp` - revert changes in these two files because they were moved to BasicCpuInfo_riscv.cpp
> 
>     * `.gitignore` - no changes needed there (you can keep those changes locally for your convenience, just not in this PR)

Done :+1:

SChernykh (approved):

SChernykh:
All good. I will soon open a new PR with everything in it (your changes and my changes) against `dev` branch.

Issue #3231, opened 2023-03-24
Haiku OS R1 / Can't compile / hwloc issue


FSOL-XDAG:
Hi everybody ! 

G33k challenge : I've try to compile XMRig 6.19.1 on Haiku OS (BeOS fork). Very nice OS if you want to discover something different than Windows & Linux.

Here are the commands I typed:

```
pkgman install cmake automake libtool autoconf
mkdir temp && cd temp
git clone https://github.com/xmrig/xmrig.git
mkdir xmrig/build && cd xmrig/scripts
./build_deps.sh && cd ../build
```

It crash on 'hwloc' : 

![image](https://user-images.githubusercontent.com/128682335/227652171-a1f8fef0-12d2-4926-a448-d2b9edc002ca.png)

Before that, I was informed about this possible issue : 

![image](https://user-images.githubusercontent.com/128682335/227651933-347e676d-c040-4ab2-a331-3bca9089510b.png)

On Haiku forum, someone tell me : 

> You can try to install hwloc from the depot aswell, maybe the build system allows select system provided libs instead of the in-tree version.

Please find complete log beside. 

[haiku_log.txt](https://github.com/xmrig/xmrig/files/11066898/haiku_log.txt)
Does anybody knows ? How to "install hwloc from the deposit" ? 

Thanks for your help ! Best regards.

SChernykh:
If hwloc doesn't support it, it will probably not work even if you manage to compile it. It's better to compile XMRig without hwloc - remove hwloc related lines from build_deps.sh and run `cmake .. -DWITH_HWLOC=OFF` when building XMRig.

FSOL-XDAG:
Ok, hwloc is ignore. But another problem occurent with OpenSSL. 

![image](https://user-images.githubusercontent.com/128682335/227889610-a9528ff9-0c38-4432-8057-a5309efbec72.png)

Don't you have any idea about this ?

SChernykh:
It's OpenSSL config script that fails, so it's not supported too. Remove OpenSSL from build_deps.sh and try `-DWITH_TLS=OFF` when building XMRig.

FSOL-XDAG:
First of all, I sought some help from the Haiku community. Apparently, the hwloc and openssl libs are already installed in the system, although it crashes at compile time (check 's' before their name) : 

![image](https://user-images.githubusercontent.com/128682335/227945636-57a2819c-3ddd-496a-86a9-8a7f6b5f2e92.png)

[...]

PR #3730, opened 2025-11-03
feat: initial support for haiku

user0-07161:
this PR adds initial support for the haiku OS.

large pages seem to crash the OS, atleast i cannot ctrl+c xmrig anymore, and running new applications and connecting via ssh does not work. therefore i disabled them.

note: hwloc packages from haiku don't seem to work, so you'll need to compile without hwloc.


user0-07161:
this would implement #3231

user0-07161:
sorry, recreating the pr on `dev`

user0-07161:
#3731

PR #3731, opened 2025-11-04
feat: initial haiku os support

user0-07161:
look at #3730 for further information


cmake/flags.cmake
```diff
@@ -46,6 +46,8 @@ if (CMAKE_CXX_COMPILER_ID MATCHES GNU)
         else()
             set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -static -Wl,--large-address-aware")
         endif()
+elseif(CMAKE_SYSTEM_NAME MATCHES "Haiku")
+        set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -static-libgcc")
```
SChernykh: - `-static-libstdc++` doesn't work?
- missing spaces before `elseif`
- Why does it use `MATCHES "Haiku"` and in the next file it's `STREQUAL "Haiku"`?

user0-07161: > - `-static-libstdc++` doesn't work?

yes, statically linking libstdc++ is unsupported on haiku and results in a linker error.

regarding the others, i'll fix them in a min

user0-07161:
should be resolved now

user0-07161:
thanks <3

user0-07161:
note: compiling with hwloc is possible if you install the `hwloc_devel` package (i was not aware of that package existing)

PR #3736, opened 2025-11-30
RISC-V: added vectorized dataset init

SChernykh:
Activated by setting `init-avx2` to 1 in config.json.

Tested on Orange Pi RV2 and got a **66% speedup**:

Before: dataset ready (28380 ms)
After: dataset ready (17018 ms)

`RxDataset::init` timings: 21865 ms before, 10639 ms after (105% speedup)

@Slayingripper please test this too (running the bench will require more code modifications, but you can just test normal mining with `"init-avx2": 1,`

Screenshot from 2025-11-30 18-58-54

### DIFF
--- a/cmake/randomx.cmake
+++ b/cmake/randomx.cmake
@@ -83,10 +83,13 @@ if (WITH_RANDOMX)
     elseif (XMRIG_RISCV AND CMAKE_SIZEOF_VOID_P EQUAL 8)
         list(APPEND SOURCES_CRYPTO
              src/crypto/randomx/jit_compiler_rv64_static.S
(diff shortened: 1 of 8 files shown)


SChernykh:
`RxDataset::init` timings: 21865 ms before, 10639 ms after (105% speedup)
This is the actual part that was rewritten using vector instructions.

Slayingripper:
Not sure I'm seeing much difference. But anyway good work 

image

SChernykh:
> Not sure I'm seeing much difference. But anyway good work

You should check the "dataset ready" line when starting XMRig, before and after you set `init-avx2` to 1 in config.json.

Issue #3737, opened 2025-12-02
Can we get a --start-paused argument?


divinity76:
I want a --start-paused argument where all initialization, config validation, and allocation is done, 
but it's started paused, waiting for a resume command. Can we get that? Would a PR be considered?

PR #3738, opened 2025-12-02
--start-paused start paused, wait for resume

divinity76:
does all the usual startup stuff, but pauses before mining, waiting for resume.

Rationale is that in some situations I only want to allocate hugepages at boot.

resolves https://github.com/xmrig/xmrig/issues/3737


QXIoa:
Intelligence at its finest

PR #3739, opened 2025-12-02
Nitpick Zen >=4

divinity76:
the previous check would break on Zen6 release.


ankitcartoon58-a11y:
Sorry I my New user

SChernykh:
@divinity76 This will not change anything because it checks a value in an enum, and there is no ZEN6 in that enum (yet). Once ZEN6 comes out, I will go through all places in the code where ZEN4/ZEN5 are mentioned and tweak them. Then it will be time to maybe do you suggested change.

ankitcartoon58-a11y:
Sorry

ankitcartoon58-a11y:
Sorry

PR #3740, opened 2025-12-05
RISC-V: added vectorized soft AES

SChernykh:
Activated when compiling with `-DARCH=native`

Before: 102.5 h/s, after: 107 h/s

Tested on Orange Pi RV2 and got a 2.2x speedup for soft AES itself, but overall speedup was only 4% because it's bottlenecked by memory access.

Note: `--randomx-1gb-pages` is crucial to get max hashrate on Orange Pi RV2

@Slayingripper

Screenshot from 2025-12-05 20-59-51
Screenshot from 2025-12-05 21-05-57

### DIFF
--- a/cmake/cpu.cmake
+++ b/cmake/cpu.cmake
@@ -55,6 +55,18 @@ if (XMRIG_RISCV)
     if(ARCH STREQUAL "native")
         enable_language(ASM)
 
+        try_run(RANDOMX_VECTOR_RUN_FAIL
+            RANDOMX_VECTOR_COMPILE_OK
+            ${CMAKE_CURRENT_BINARY_DIR}/
+            ${CMAKE_CURRENT_SOURCE_DIR}/src/crypto/randomx/tests/riscv64_vector.s
+            COMPILE_DEFINITIONS "-march=rv64gcv_zicbop")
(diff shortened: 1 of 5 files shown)


Slayingripper:
I've switched to using to an NVME , the biggest difference I've seen is  start up time when running xmrig.  I'll see if i can edit the 
image

I've enabled huge pages by also editig the uboot config file  added this line: 

```
extraargs=default_hugepagesz=1G hugepagesz=1G hugepages=3
```
image

I have some other stuff running on this machine, so it's not completely clean as I use it for other things as well  (Docker). But any enhancement is always welcome.

SChernykh:
1gb pages worked for me just by using sudo and `--randomx-1gb-pages` in the command line, but I don't run anything on that board. Also, having copper heatsinks on CPU and memory helps a lot with hashrate and stability.

Slayingripper:
> 1gb pages worked for me just by using sudo and `--randomx-1gb-pages` in the command line, but I don't run anything on that board. Also, having copper heatsinks on CPU and memory helps a lot with hashrate and stability.

I also have a heatsink with a fan. But in any case, improvement is always good.

PR #3742, opened 2025-12-07
Add support for Juno Cash

kobiyagi:
Ive created a document outlining *some* modifications needed to add support for Juno Cash (https://juno.cash/).


user0-07161:
why are you adding docs instead of directly implementing it?

SChernykh:
I don't think it's the right place for this kind of documentation. Also, it looks like AI-generated.

PR #3725, opened 2025-10-22
RISC-V integration and JIT compiler

SChernykh:
- Groundwork was done by @Slayingripper in #3724
- JIT compiler ported by me from https://github.com/tevador/RandomX/
- Added new cmake options: `-DARCH=native` for RISC-V builds - it will test and enable `zba` and `zbb` extenstions for RandomX

Closes #1924

PR #3746, opened 2025-12-26
RISC-V: vectorized RandomX main loop

SChernykh:
This hasn't been tested on a real hardware yet, but it works in QEMU.

@Slayingripper if you have time, please test this PR against the latest release. Both must be built with
````
cmake .. -DCMAKE_BUILD_TYPE=Release -DARCH=native
````

Vector version should be faster, but I don't know by how much.

### DIFF
--- a/cmake/cpu.cmake
+++ b/cmake/cpu.cmake
@@ -51,42 +51,74 @@ if (XMRIG_RISCV)
     # default build uses the RV64GC baseline
     set(RVARCH "rv64gc")
 
+    enable_language(ASM)
+
+    try_run(RANDOMX_VECTOR_RUN_FAIL
+        RANDOMX_VECTOR_COMPILE_OK
+        ${CMAKE_CURRENT_BINARY_DIR}/
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/crypto/randomx/tests/riscv64_vector.s
+        COMPILE_DEFINITIONS "-march=rv64gcv")
+
+    if (RANDOMX_VECTOR_COMPILE_OK AND NOT RANDOMX_VECTOR_RUN_FAIL)
+        set(RVARCH_V ON)
+        message(STATUS "RISC-V vector extension detected")
+    else()
+        set(RVARCH_V OFF)
+    endif()
+
+    try_run(RANDOMX_ZICBOP_RUN_FAIL
+        RANDOMX_ZICBOP_COMPILE_OK
+        ${CMAKE_CURRENT_BINARY_DIR}/
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/crypto/randomx/tests/riscv64_zicbop.s
+        COMPILE_DEFINITIONS "-march=rv64gc_zicbop")
+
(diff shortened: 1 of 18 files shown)


SChernykh:
Finally did a test, got an improvement from 107.4 to 110.1 H/s.
Screenshot from 2026-01-04 18-02-55

Slayingripper:
I'm not sure why we have such a big difference in numbers. The only thing running on my PI is PiHole 
image

SChernykh:
You mistyped (missed `M`), need to type `sudo ./xmrig --bench=1M --randomx-1gb-pages`

Slayingripper:
image

SChernykh:
I don't know why it's slower for you. My RV2 has only heatsinks but no fan like on yours, so your RV2 is definitely not overheating. But you had 101.8 h/s all the way back here: https://github.com/xmrig/xmrig/pull/3725#issuecomment-3433678228
Maybe OS reinstall will help you.

Slayingripper:
I'll explain what I've done. I've deleted the repository, redownloaded it, switched to this pr like so ` gh pr checkout 3746 `.  Then just proceeded as normal, `mkdir build` &#8594; `cd build/`&#8594; ` cmake .. -DCMAKE_BUILD_TYPE=Release -DARCH=native ` &#8594; ` make -j8` &#8594; `sudo ./xmrig --bench=1M --randomx-1gb-pages `. 

Am I missing something? 


> I don't know why it's slower for you. My RV2 has only heatsinks but no fan like on yours, so your RV2 is definitely not overheating. But you had 101.8 h/s all the way back here: [#3725 (comment)](https://github.com/xmrig/xmrig/pull/3725#issuecomment-3433678228) Maybe OS reinstall will help you.

SChernykh:
Everything looks correct. I think either something is using a lot of CPU in background, or you set power governor to "power saving" mode or something like this.

SChernykh:
@Slayingripper I now have the same problem. Hashrate dropped to 85 H/s and I can't do anything to get it back to 110 H/s. `--randomx-1gb-pages` doesn't help, in fact hashrate stays the same as with the regular huge pages. So I suspect that 1 GB huge pages stopped working, and OS is allocating the regular huge pages in the background.

Slayingripper:
> @Slayingripper I now have the same problem. Hashrate dropped to 85 H/s and I can't do anything to get it back to 110 H/s. `--randomx-1gb-pages` doesn't help, in fact hashrate stays the same as with the regular huge pages. So I suspect that 1 GB huge pages stopped working, and OS is allocating the regular huge pages in the background.

At least I now know I didn't do anything on purpose. I tried disabling everything (Docker, PiHole, etc) to see if it would do anything, but nothing happened. Maybe Huge pages is something we should look into. I don't think this is a RISC-V problem, more likely an OS issue.

SChernykh:
@Slayingripper I don't know what changed, but now I get 118.3 h/s with the latest XMRig. It seems that it's random between reboots, and I think it depends on which parts of physical memory get allocated for the dataset:
image

PR #3762, opened 2026-01-20
Fix keepalive timer logic

HashVault:
Reset keepalive timer on send instead of receive - the pool needs to know the client is alive, not the other way around.

Also remove `m_keepAlive = 0` in `ping()` which was disabling the timer after the first ping, so only one keepalive would ever be sent.


xmrig:
Merged into the `dev` branch. This PR will be closed automatically when the `dev` branch is merged into `master`.

Thank you.

PR #3766, opened 2026-01-21
feat: stability & hardening improvements

oxyzenQ:
Bounded/defensive behavior to avoid unbounded growth & risky teardown paths
Fail-fast on oversized stratum lines (disconnect cleanly to avoid protocol desync)
Avoid abort on NetBuffer pool teardown with outstanding allocations (prefer safe leak over crash)
Optional GPU host-loop throttling via per-thread yield_sleep (default 0, no behavior change)
Added sanitizer presets in CMakePresets.json (ASan/UBSan/TSan)

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,62 @@
+{
+  "version": 3,
+  "cmakeMinimumRequired": {
+    "major": 3,
+    "minor": 19,
+    "patch": 0
+  },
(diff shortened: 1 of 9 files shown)


oxyzenQ:
Closing this PR and resubmitting as two separate PRs targeting dev as requested. Thanks for the guidance.

PR #3771, opened 2026-01-30
 Intel CPU and RandomX Optimizations

h9zdev:
This set of changes introduces performance and thermal efficiency optimizations specifically for Intel mobile CPUs (Comet Lake and similar). It addresses common bottlenecks like L3 cache thrashing and aggressive power management throttling. Additionally, it provides an experimental prefetchw mode and unlocks the 0% donation level.

### DIFF
--- a/src/backend/cpu/CpuWorker.cpp
+++ b/src/backend/cpu/CpuWorker.cpp
@@ -94,7 +94,11 @@ xmrig::CpuWorker<N>::CpuWorker(size_t id, const CpuLaunchData &data) :
     else
 #   endif
     {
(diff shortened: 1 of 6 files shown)


SChernykh:
Almost all of these changes make no sense and actually don't help with the "cache thrashing". Shifting scratchpad in increments of 32 KB certainly doesn't. `codePosFirst` change will most likely make it crash instead of mining faster. MSR tweak change to make it more power hungry to hash faster is not the right thing to do. It's up to the user to tweak CPUs power setting in the OS to achieve what is needed from the specific system.

TLDR: AI slop pull request - most likely will crash, at best the performance will be the same (not including the MSR tweak).

PR #3796, opened 2026-03-30
RISC-V RandomX: Enable hardware AES and optimize memory prefetching

oxyzenQ:
This commit addresses performance bottlenecks in the RandomX implementation for RISC-V architecture by enabling hardware AES support when available and adding software prefetching to reduce memory latency.

Changes:

src/crypto/rx/RxVm.cpp:
- Remove forced software AES on RISC-V (was unconditionally blocking HW AES)
- Enable RANDOMX_FLAG_HARD_AES when CPU has AES support (Zvkned extension)
- Simplify code by removing #ifndef XMRIG_RISCV guard that prevented hardware acceleration even when available

src/crypto/randomx/aes_hash_rv64_vector.cpp:
- Add #include "intrin_portable.h" for rx_prefetch_* macros
- Add PREFETCH_DISTANCE (7168 bytes) matching x86 implementation
- Insert rx_prefetch_t0() calls for upcoming scratchpad blocks during hashAndFillAes1Rx4_RVV() loop processing
- Split loop into prefetch-capable section and tail section for remaining data to avoid prefetching past buffer bounds

src/crypto/randomx/dataset.cpp:
- Restructure initDatasetItem() prefetch logic: prefetch first mix block before loop entry
- Move prefetch of next iteration's data to end of current iteration, overlapping superscalar execution with memory fetch
- Eliminates prefetch-then-immediately-use pattern that didn't hide latency

Impact:
- RISC-V systems with Zvkned extension now use hardware AES instead of software fallback (significant throughput improvement)
- RVV soft AES path now overlaps memory access with computation via prefetching, reducing pipeline stalls
- Dataset initialization overlaps superscalar execution with cache line fetches, improving parallelism

Tested: Build verified with cmake + make -j4

### DIFF
--- a/src/crypto/randomx/aes_hash_rv64_vector.cpp
+++ b/src/crypto/randomx/aes_hash_rv64_vector.cpp
@@ -31,6 +31,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include "crypto/randomx/soft_aes.h"
 #include "crypto/randomx/randomx.h"
+#include "crypto/randomx/intrin_portable.h"
 
 static FORCE_INLINE vuint32m1_t softaes_vector_double(
 	vuint32m1_t in,
@@ -229,6 +230,10 @@ void hashAndFillAes1Rx4_RVV(void *scratchpad, size_t scratchpadSize, void *hash,
 	const vuint8m1_t& lutdec_index2 = lutenc_index2;
 	const vuint8m1_t lutdec_index3 = __riscv_vle8_v_u8m1(lutDecIndex[3], 32);
 
+	constexpr int PREFETCH_DISTANCE = 7168;
+	const char* prefetchPtr = ((const char*)scratchpad) + PREFETCH_DISTANCE;
+	scratchpadEnd -= PREFETCH_DISTANCE;
+
 	//process 64 bytes at a time in 4 lanes
 	while (scratchpadPtr < scratchpadEnd) {
 #define HASH_STATE(k) \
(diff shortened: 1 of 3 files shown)


src/crypto/randomx/aes_hash_rv64_vector.cpp
```diff
@@ -247,6 +252,22 @@ void hashAndFillAes1Rx4_RVV(void *scratchpad, size_t scratchpadSize, void *hash,
...
+
+	// Process remaining data without prefetching
+	scratchpadEnd += PREFETCH_DISTANCE;
+	while (scratchpadPtr < scratchpadEnd) {
+		HASH_STATE(0);
+		HASH_STATE(1);
+
+		FILL_STATE(0);
+		FILL_STATE(1);
+
```
SChernykh, line 270: This prefetch doesn't have any effect, I tested it on the real RISC-V board. It only makes code more complex.

src/crypto/randomx/dataset.cpp
```diff
@@ -154,6 +157,12 @@ namespace randomx {
 				rl[q] ^= load64_native(mixBlock + 8 * q);
 
 			registerValue = rl[prog.getAddressRegister()];
+
+			// Prefetch next mix block before continuing (if not last iteration)
+			if (i + 1 < RandomX_CurrentConfig.CacheAccesses) {
+				mixBlock = getMixBlock(registerValue, cache->memory);
+				rx_prefetch_nta(mixBlock);
+			}
```
SChernykh, line 165: This changes literally nothing. Both versions of the code do exactly the required number of prefetches, when they are required. Only this code is more complex. Unnecessary.

src/crypto/rx/RxVm.cpp
```diff
@@ -29,17 +29,11 @@ randomx_vm *xmrig::RxVm::create(RxDataset *dataset, uint8_t *scratchpad, bool so
...
 
-    // On RISC-V, force software AES path even if CPU reports AES capability.
-    // The RandomX portable intrinsics will throw at runtime when HAVE_AES is not defined
-    // for this architecture. Until native AES intrinsics are wired for RISC-V, avoid
-    // setting HARD_AES to prevent "Platform doesn't support hardware AES" aborts.
-#   ifndef XMRIG_RISCV
-    if (!softAes) {
+    // On RISC-V, use hardware AES (Zvkned) when available via hasAES() check.
+    // Software AES is only used as fallback when hardware support is not present.
+    if (!softAes && Cpu::info()->hasAES()) {
```
SChernykh, line 34: This is the only change that makes sense in this PR. The other two are unneeded.

SChernykh (changes requested):

src/crypto/rx/RxVm.cpp
oxyzenQ: if this oke. if bad do not merge this PR this is very bad idea.

src/crypto/randomx/dataset.cpp
oxyzenQ: oh im very stupid. do not merge this PR this is very bad idea.

src/crypto/randomx/aes_hash_rv64_vector.cpp
oxyzenQ: oh zxd. do not merge this PR this is very bad idea.

PR #3801, opened 2026-04-14
Fixed GPU validation for `rx/2`.

im10furry:
Required for `rx/2` GPU support.

  `JobResults::getResults()` revalidates GPU nonces on CPU before submission. For `rx/2`, the final
  share target must be checked against the commitment, not the raw RandomX hash.

  This patch updates the GPU verification path to:
  - calculate the raw RandomX hash for the returned nonce
  - calculate the `rx/2` commitment from the blob and raw hash
  - validate the commitment against the job target
  - pass the raw hash as extra data so the submit path can send the correct `commitment` field

  No behavior changes for other algorithms.


SChernykh:
`rx/2` is not supported on GPU due to its uselessness. `rx/0` GPU support was just a proof of concept to get the real performance numbers.

PR #3820, opened 2026-05-25
ARM64 RandomX JIT: dataset prefetch + non-temporal loads (+~8% on M4 base)

aa022:
Two Apple-silicon-targeted tweaks to the aarch64 RandomX JIT:

- Early dataset prefetch: when readReg2/readReg3 are finalized well before the end of the program body, emit the next iteration's dataset-line prefetch early to hide more DRAM latency on the serial scalar chain.
- Non-temporal dataset loads: each 64-byte dataset line is read once and never reused, so ldp -> ldnp avoids evicting the hot scratchpad, and the prefetch hint moves pldl2strm -> pldl1strm to match the longer lead time.

Measured ~8% hashrate gain on Apple M4 base over dev (7eadfdc9)
                  ~22% hashrate gain on Apple M1 air over dev (7eadfdc9)

m1-air-bench

### DIFF
--- a/src/crypto/randomx/jit_compiler_a64.cpp
+++ b/src/crypto/randomx/jit_compiler_a64.cpp
@@ -152,20 +152,6 @@ void JitCompilerA64::generateProgram(Program& program, ProgramConfiguration& con
 		(this->*engine[instr.opcode])(instr, codePos);
 	}
 
-	{
-		const uint32_t rr2Off = reg_changed_offset[config.readReg2];
-		const uint32_t rr3Off = reg_changed_offset[config.readReg3];
-		const uint32_t maxOff = (rr2Off > rr3Off) ? rr2Off : rr3Off;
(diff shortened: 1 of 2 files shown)


SChernykh:
> Early dataset prefetch: when readReg2/readReg3 are finalized well before the end of the program body, emit the next iteration's dataset-line prefetch early to hide more DRAM latency on the serial scalar chain.

I think this change does nothing because dataset prefetch is moved just a few instructions earlier in the code. And the second dataset prefetch from the same location is still there, so maybe it even slows down the code a little.

The speedup you got is from `pldl1strm` and `ldnp` only. Try to test without the early prefetch change.

aa022:
yeah you are right, 4430.4 H/s with the prefetch on 1M bench, 4431.9 H/s without; its just a redundant prefetch

SChernykh (approved):

PR #3827, opened 2026-06-30
Implement support for PowerPC (PPC/PPC64/PPC64BE)

PalindromicBreadLoaf:
This adds in support for PowerPC devices (including 32-bit PPC, 64-bit PPC little endian, and 64-bit PPC big endian [such as Power8 and Power9). Whilst I don't have a device to test PPC64, a VM did show a successful run, despite being incredibly slow. On hardware, I tested on a Nintendo WiiU running Gentoo Linux and got what you'd expect performance-wise. I completely understand if this is outside the scope of this project to support such an esoteric architecture, just wanted to throw this here in case it was useful.


SChernykh:
It is a bit out of scope here, because no one realistically mines on PPC. But full PPC support would be nice in the upstream RandomX repository. You should cooperate with the author of this PR: https://github.com/tevador/RandomX/pull/320

zachary17556:
this will be perfect for xbox360 and for the ps3

PalindromicBreadLoaf:
> this will be perfect for xbox360 and for the ps3

As funny as it would be, I don't think the Xbox360 has enough cache to run the miner, although I think PS3 has 2MB on it's SPUs, but I don't know if we can access that in PS3 Linux or not. I was lucky the WiiU had one core with 2MB of cache.

PR #3831, opened 2026-07-17
feat: Add Tari (RandomX-Tari) daemon mining support

PotR3288:
# Add Tari (RandomX-Tari) daemon mining support

## Description

Add support for mining on the [Tari](https://tari.com/) network via its daemon protocol to base Tari nodes with xmrig_proxy enabled. Tari uses a 76-byte RandomX blob format that differs from standard Monero block templates, requiring dedicated parsing, hashing, and difficulty validation paths.

### Algorithm & Coin registration

- src/base/crypto/Algorithm.{cpp,h} — Register RX_TARI (id 0x72151274, alias "rx/tari", "randomxt"), following the existing RandomX fork naming convention (rx/<coin>)
- src/base/crypto/Coin.cpp — Add coin info entry: ticker XTM, 120s block time, 12-digit precision

### RandomX configuration & hashing

- src/crypto/randomx/randomx.{cpp,h} — Declare RandomX_ConfigurationTari using default base config (3 Argon iterations, standard flags)
- src/crypto/rx/RxAlgo.cpp — Route RX_TARI to RandomX_TariConfig in the algo dispatcher
- src/crypto/rx/RxTari.{cpp,h} — New module: tari_randomx_calculate_hash() hashes the full 76-byte blob directly without Monero-format repacking
- cmake/randomx.cmake — Include new source/header files in build

### Block template parsing

- src/base/tools/cryptonote/BlockTemplate.{cpp,h} — Dedicated Tari path: parse 76-byte blob (3-byte header + 32B mining hash + 8B nonce LE + 33B pow data), expose isTari(), tariMiningHash(), tariNonce(), tariPowData() accessors

### Daemon client integration

- src/base/net/stratum/DaemonClient.{cpp,h} — Parse Tari-specific fields from getblocktemplate: min_nonce/max_nonce (u64 range), difficulty (u64 target). Retain last successful blocktemplate to avoid redundant requests. Log Tari difficulty as 2^64 / diff for readability
- src/base/net/stratum/Job.{cpp,h} — 8-byte nonce size for RX_TARI. Store daemon-assigned min_nonce/max_nonce range. Implement achievesDifficulty() using multi-precision multiplication (scalar \* targetDiff <= U256::MAX) with MSVC \_umul128 and GCC/clang unsigned \__int128 support
- src/base/net/stratum/SelfSelectClient.cpp — Adjust self-select submission logging for Tari (omit redundant comparison diff, as difficulty and achieved will always match in this case)

### CPU worker hash loop

- src/backend/cpu/CpuWorker.cpp — Early-exit Tari path before generic RandomX: direct tari_randomx_calculate_hash() call, enforce daemon-assigned nonce boundary, increment local nonce, skip pipelined nextRound(). On job consumption, align per-worker starting nonces to daemon range using worker_id \* N \* kReserveCount offset

### Job results & submission

- src/net/JobResult.h — actualDiff() uses full 256-bit difficulty formula for Tari (not truncated low-32-bits)
- src/net/JobResults.{cpp,h} — Widened submit() overloads from uint32_t to uint64_t nonce parameter so 8-byte Tari nonces pass through without truncation
- src/base/net/stratum/Client.cpp — Log submitted nonce as 16-digit hex for Tari, 8-digit for all other algorithms

## Backward compatibility

All changes are gated behind XMRIG_ALGO_RANDOMX / Algorithm::RX_TARI checks. Mining Monero or any existing algorithm is unaffected. No configuration defaults were modified.

## Testing

- Tested solo single- and multi-miner mining against local mainnet Tari nodes
- Verified: blocks found, random nonce range assignment across workers, correct 256-bit difficulty validation
- Both GCC/Clang and MSVC compilation paths verified (no unsigned \__int128 in MSVC path)

_**No address validation of Tari addresses. Use caution when entering payment address**_

### DIFF
--- a/cmake/randomx.cmake
+++ b/cmake/randomx.cmake
@@ -26,6 +26,7 @@ if (WITH_RANDOMX)
         src/crypto/rx/RxQueue.h
         src/crypto/rx/RxSeed.h
         src/crypto/rx/RxVm.h
+        src/crypto/rx/RxTari.h
     )
 
     list(APPEND SOURCES_CRYPTO
@@ -54,6 +55,7 @@ if (WITH_RANDOMX)
         src/crypto/rx/RxDataset.cpp
         src/crypto/rx/RxQueue.cpp
         src/crypto/rx/RxVm.cpp
+        src/crypto/rx/RxTari.cpp
     )
 
     if (WITH_ASM AND CMAKE_C_COMPILER_ID MATCHES MSVC)
--- a/src/backend/cpu/CpuWorker.cpp
+++ b/src/backend/cpu/CpuWorker.cpp
@@ -21,6 +21,7 @@
 #include <mutex>
 
 
+#include "base/io/log/Log.h"
 #include "backend/cpu/Cpu.h"
 #include "backend/cpu/CpuWorker.h"
 #include "base/tools/Alignment.h"
@@ -35,6 +36,7 @@
 #include "crypto/rx/Rx.h"
 #include "crypto/rx/RxCache.h"
 #include "crypto/rx/RxDataset.h"
+#include "crypto/rx/RxTari.h"
 #include "crypto/rx/RxVm.h"
 #include "crypto/ghostrider/ghostrider.h"
 #include "net/JobResults.h"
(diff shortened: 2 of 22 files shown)


SChernykh:
This PR and code looks 100% AI generated, and whatever LLM you used went the wrong way in implementing it. It shouldn't be that many files edited and that much "Tari-specific" code. `tari_randomx_calculate_hash` is a cherry on top of what shouldn't have been done.

PotR3288:
Thanks for looking at this. My intent was to avoid affecting any other aspect of xmrig's operation in order to add Tari support. I will close this PR and re-submit once I have a better handle on the appropriate approach.

PR #3832, opened 2026-07-17
    Fix NUMA memory binding on Windows with >64 CPUs (processor groups)

PotR3288:
fix: multi-step NUMA memory binding fallback for Windows >64 CPUs
      
 On Windows with more than 64 logical processors, hwloc's membind functions
 can fail silently because they internally call SetThreadAffinityMask(), which
 only accepts a 64-bit mask (one processor group).
 
This PR adds a fallback cascade: nodeset membind → cpuset membind → cpubind via
SetThreadGroupAffinity() → thread affinity.

Also adds SetThreadGroupAffinity support in Platform_win.cpp for CPU >= 64,
and a hwloc_const_nodeset_t typedef needed by the new function signature.

Files changed:
- src/backend/cpu/interfaces/ICpuInfo.h — add hwloc_const_nodeset_t typedef
- src/backend/cpu/platform/HwlocCpuInfo.cpp — multi-step do_membind() fallback cascade
- src/base/kernel/Platform_win.cpp — SetThreadGroupAffinity for CPU >= 64
      
Backward Compatibility

- Linux/macOS: No change — direct nodeset binding succeeds wherever the old path did
- Windows <64 CPUs: No change — first fallback always succeeds
- Windows 64+ CPUs with hwloc: Fixed — correct NUMA binding on all processor groups
- Windows 64+ CPUs without hwloc: Fixed — threads correctly placed in higher groups

Testing
- Compared stock v6.26.0 with a build including these modifications on Windows 11 systems with >64 threads
- Without the adjustments, thread memory failed to bind and datasets failed to load
- The build with the changes ran normally. Datasets properly loaded and mining hashrate was as expected for the tested maximum of 192 threads
- Linux build ran as before the modifications


.qwen/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(dir *)"
+    ]
+  }
+}
```
SChernykh: Please clean up the LLM stuff.

ProcessorGroupChanges.md
```diff
@@ -0,0 +1,204 @@
+# Processor Groups Fix — Changes Explained
```
SChernykh: This whole file is not needed here.

ProcessorGroupFix.md
```diff
@@ -0,0 +1,364 @@
+# Windows Processor Groups Fix for XMRIG NUMA Binding
```
SChernykh: Same, not needed.

src/backend/cpu/interfaces/ICpuInfo.h
```diff
@@ -128,7 +129,8 @@ class ICpuInfo
     virtual uint32_t model() const                                                  = 0;
 
 #   ifdef XMRIG_FEATURE_HWLOC
-    virtual bool membind(hwloc_const_bitmap_t nodeset)                              = 0;
+    virtual bool membind(hwloc_const_bitmap_t cpuset)                               = 0;
+    virtual bool membind_nodeset(hwloc_const_nodeset_t nodeset)                     = 0;
```
SChernykh: Why two functions? One function must handle everything properly, and not leave it to the caller to figure out all the edge cases.

src/backend/cpu/platform/HwlocCpuInfo.h
```diff
@@ -38,7 +38,8 @@ class HwlocCpuInfo : public BasicCpuInfo
     ~HwlocCpuInfo() override;
 
 protected:
-    bool membind(hwloc_const_bitmap_t nodeset) override;
+    bool membind(hwloc_const_bitmap_t cpuset) override;
+    bool membind_nodeset(hwloc_const_nodeset_t nodeset) override;
```
SChernykh: Again, it should be one function.

src/crypto/rx/RxNUMAStorage.cpp
```diff
@@ -49,12 +49,28 @@ static bool bindToNUMANode(uint32_t nodeId)
         return false;
     }
 
-    if (Cpu::info()->membind(node->nodeset)) {
+    // Try direct nodeset binding first (avoids cpuset round-trip on Windows processor groups)
+    if (Cpu::info()->membind_nodeset(node->nodeset)) {
```
SChernykh: And this file shouldn't be changed at all just because you introduced two `membind` functions.

SChernykh (changes requested):

src/backend/cpu/platform/HwlocCpuInfo.cpp
```diff
@@ -191,17 +192,87 @@ xmrig::HwlocCpuInfo::~HwlocCpuInfo()
 }
 
 
-bool xmrig::HwlocCpuInfo::membind(hwloc_const_bitmap_t nodeset)
+static bool do_membind(hwloc_topology_t topology, hwloc_const_bitmap_t cpuset, hwloc_const_nodeset_t nodeset)
```
SChernykh: Why did you make it a separate function, which is also `static` and called from a single place in the code? Also always with `cpuset == nullptr`, so everything that works with `cpuset` below is a dead code now. Feels like something incomplete.

PotR3288: Aye, it was incomplete. I don't know git very well yet, this should have been WIP

PotR3288:
Compiled and tested on: 

- Windows >64-CPU with and without hwloc
- Windows <64-CPU with and without hwloc
- Linux
- Mac - ARM64

PR #3836, opened 2026-07-29
Added TKM

71mengr:
Kindly add TKM

### DIFF
--- a/src/backend/cpu/CpuWorker.cpp
+++ b/src/backend/cpu/CpuWorker.cpp
@@ -363,20 +363,22 @@ void xmrig::CpuWorker<N>::start()
                     }
                     else
 #                   endif
+                    {
+                        const bool accepted = job.algorithm() == Algorithm::RX_TKM ? job.isTkmHashAccepted(m_hash + (i * 32)) : value < job.target();
+                        if (accepted) {
+                            uint8_t* extra_data = nullptr;
 
-                    if (value < job.target()) {
-                        uint8_t* extra_data = nullptr;
+                            if (job.algorithm().family() == Algorithm::RANDOM_X) {
+                                if (RandomX_CurrentConfig.Tweak_V2_COMMITMENT) {
+                                    extra_data = m_commitment;
+                                }
+                                else if (job.hasMinerSignature()) {
+                                    extra_data = miner_signature_saved;
+                                }
+                            }
 
-                        if (job.algorithm().family() == Algorithm::RANDOM_X) {
-                            if (RandomX_CurrentConfig.Tweak_V2_COMMITMENT) {
-                                extra_data = m_commitment;
-                            }
-                            else if (job.hasMinerSignature()) {
-                                extra_data = miner_signature_saved;
-                            }
+                            JobResults::submit(job, current_job_nonces[i], m_hash + (i * 32), extra_data);
                         }
-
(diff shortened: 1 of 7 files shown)


SChernykh:
This coin is completely new and not listed anywhere - I can't even find it on any market cap/mining websites. As a general rule, XMRig adds only well-established coins, otherwise it would have to add dozens of coins every year.

71mengr:
> This coin is completely new and not listed anywhere - I can't even find it on any market cap/mining websites. As a general rule, XMRig adds only well-established coins, otherwise it would have to add dozens of coins every year.

Website: https://tkmchain.site,  https://pool.tkmchain.site,  https://block.tkmchain.site.

SChernykh:
I would recommend to make the Stratum compatible instead (can be done on the pool side), and the target check compatible as well (can be done with a fork - easy to do while the coin is still new). Then you can just use stock XMRig and rx/0 algorithm.

71mengr:
> I would recommend to make the Stratum compatible instead (can be done on the pool side), and the target check compatible as well (can be done with a fork - easy to do while the coin is still new). Then you can just use stock XMRig and rx/0 algorithm.

Why A Pool-Only Stratum Update Cannot Support Stock XMRig

## Short Answer

Updating only the pool will not make unmodified stock XMRig compatible with TKMChain's current mining format.           
The pool controls job delivery and share submission, but the miner controls which bytes are hashed, where the nonce is written, and how a candidate hash is compared with the target. TKMChain currently differs from stock XMRig in all three areas.             
## Current TKMChain Work Format            
TKMChain currently calculates RandomX over a 40-byte input:

```text
RandomX(seedHash, sealHash[32] || headerNonce[8])
```

The node constructs that input in `go-tkmchain/consensus/randomx/randomx.go`:

```text
bytes 0-31   = TKMChain seal hash
bytes 32-39  = eight-byte block-header nonce
```

The node interprets the resulting 32-byte RandomX hash as a big-endian integer and accepts it when:

```text
hash <= (2^256 - 1) / difficulty
```

## What The Patched TKM XMRig Changes

The miner in `xmrig` adds the `rx/tkm` algorithm. That patch is not merely a different algorithm name. It changes behavior needed to match TKMChain:

- accepts TKMChain's 40-byte `sealHash || nonce` blob;
- uses nonce offset `36` for the four bytes varied by XMRig;
- hashes the exact TKMChain blob with the supplied RandomX seed;
- compares the complete 32-byte output with TKMChain's big-endian 256-bit target;
- submits the result through the TKM-compatible Stratum job and share format.

The pool extends the submitted four-byte miner nonce into the eight-byte header nonce expected by the node.

## What Stock XMRig `rx/0` Expects

Unmodified stock XMRig uses its standard CryptoNote/Monero-style RandomX job behavior. In particular, it normally:

- expects a CryptoNote-compatible mining blob;
- writes a four-byte nonce at offset `39`;
- requires a blob long enough to contain that nonce;
- hashes the complete blob it received;
- uses the standard XMRig target interpretation and local share check.

Those rules do not match TKMChain's current 40-byte blob and target comparison

SChernykh:
This is why I also mentioned a fork.

71mengr:
> This is why I also mentioned a fork.

Not bad if you test it https://github.com/tkmchain/xmrig/releases/tag/v6.26.0-tkm.8 and our pool https://pool.tkmchain.site/

PR #3839, opened 2026-08-19
RISC-V: detect VLEN at runtime and speed up the RVV paths

KiritakeKumi:
Every RVV path now checks the vector length of the calling thread instead of assuming it is at least 256 bits, which also fixes silently wrong hashes on narrower vector units. VLEN is read per thread because it is a per-hart property and need not be uniform across a system ( Take the SpacemiT K3, for example; this chip features two clusters, but they differ in VLEN. ) .

The vectorized dataset init scales with the actual vector length instead of a fixed four lanes, and the RVV AES kernels no longer use indexed loads and stores.

Also fixes an out-of-bounds write in the vectorized dataset init when the item count is not a multiple of the lane count.

Measured on a SpacemiT K3 (rx/0, --bench=250K), hash sums unchanged:

  X100 (VLEN 256):   518.4 -> 536.8 H/s
  A100 (VLEN 1024):  186.1 -> 229.1 H/s, dataset init 29.8 s -> 15.5 s

### DIFF
--- a/src/Summary.cpp
+++ b/src/Summary.cpp
@@ -42,6 +42,11 @@
 #endif
 
 
+#ifdef XMRIG_RISCV
+#   include "backend/cpu/platform/riscv_vlen.h"
+#endif
+
+
 namespace xmrig {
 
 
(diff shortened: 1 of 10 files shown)


SChernykh:
This is interesting. All RISC-V hardware is very different. I'll need to test it on Orange Pi RV2 to see how it reacts to this new code.

SChernykh:
Orange Pi RV2 (1 thread, to reduce dependency on RAM):
- Dataset init - before 14839 ms, after 15578 ms (5% slower)
- Hashrate - before 28.4-29.24 h/s, after 28.81-29.55 h/s (no meaningful difference, hashrate measurement is unstable)

Based on this test, it makes sense to have AES changes in a separate PR because they improve hashrate on SpacemiT K3 and don't change anything on Ky X1 (maybe a tiny improvement), so overall it's a net positive.

Dataset changes are controversial.

KiritakeKumi:
> Orange Pi RV2 (1 thread, to reduce dependency on RAM):
> 
> * Dataset init - before 14839 ms, after 15578 ms (5% slower)
> * Hashrate - before 28.4-29.24 h/s, after 28.81-29.55 h/s (no meaningful difference, hashrate measurement is unstable)
> 
> Based on this test, it makes sense to have AES changes in a separate PR because they improve hashrate on SpacemiT K3 and don't change anything on Ky X1 (maybe a tiny improvement), so overall it's a net positive.
> 
> Dataset changes are controversial.

Okay, I will break it down and continue investigating further processing for the dataset.

PR #3841, opened 2026-08-22
RISC-V: avoid indexed loads/stores in the RVV AES kernels

KiritakeKumi:
The four 16-byte AES lanes are consecutive in memory but the kernels need them grouped as {0,2} and {1,3}, which the strided AES_HASH_STRIDE_X2 index vector did in a single indexed access. Indexed accesses turn out to be an order of magnitude slower than unit-stride ones on both SpacemiT cores (SpacemiT K3 chip) , so pick the two halves up separately and glue them together with a slide.

Measured with 8 independent destination registers, AVL=8, e32/LMUL=1:

X100 (VLEN 256)   A100 (VLEN 1024)
  vluxei32 (gather)          7.28 ns          129.8 ns
  vsuxei32 (scatter)         7.28 ns           59.7 ns
  vle32  (unit-stride)       0.46 ns            9.46 ns
  vse32  (unit-stride)       0.91 ns            1.67 ns
  vslideup.vi                0.91 ns            4.45 ns

Applied to the hot loops of all four kernels. The remaining indexed accesses run once per call on a 64-byte state buffer and are amortized over a 2 MB fill, so they are left alone.

hashAndFillAes1Rx4 throughput on a SpacemiT K3:

  X100 (VLEN 256):   2.16 -> 4.62 GB/s  (2.14x)
  A100 (VLEN 1024):  0.21 -> 0.64 GB/s  (3.05x)

End to end, --bench=250K -t 8 on the K3 X100 cluster (VLEN 256), two runs of each build, alternating order:

  before   594.1, 593.4 h/s
  after    620.2, 621.6 h/s   (+4.6%)
  
This PR is a split-out commit from https://github.com/xmrig/xmrig/pull/3839.
cc @SChernykh

### DIFF
--- a/src/crypto/randomx/aes_hash_rv64_vector.cpp
+++ b/src/crypto/randomx/aes_hash_rv64_vector.cpp
@@ -69,6 +69,26 @@ static constexpr uint32_t AES_HASH_1R_XKEY11[8] = { 0x61b263d1, 0x51f4e03c, 0xee
 static constexpr uint32_t AES_HASH_STRIDE_X2[8] = { 0, 4, 8, 12, 32, 36, 40, 44 };
 static constexpr uint32_t AES_HASH_STRIDE_X4[8] = { 0, 4, 8, 12, 64, 68, 72, 76 };
 
+/*
+	The four 16-byte AES lanes sit consecutively in memory, but the kernels need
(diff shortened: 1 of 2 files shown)


SChernykh (approved):
It passes the tests on my RV2, and looks to be a little bit faster.

